PR #18, opened 2026-04-15
Docs/architecture

Imbad0202:


Imbad0202:
Overall Review

  Request changes

  This is a useful document, but I don't think it is merge-ready yet because several parts
  currently diverge from the repo's canonical behavior and contracts.

  The main issues are:
  1. it redefines the checkpoint model by describing only 8 human gates and treating Stages
  2.5 / 4.5 as machine-only, while the current pipeline docs still require explicit user
  confirmation after every completed stage;
  2. the Stage 2.5 row contains factual mismatches against the current integrity protocol
  (both the 7-mode taxonomy labels and the "100% verified" claim mapping);
  3. the Stage 2 / `academic-paper` row compresses the skill into a 10-agent model even though
  the current skill is documented as a 12-agent pipeline;
  4. the flow diagram omits the currently documented review-to-revision coaching transitions;
  5. the data-isolation section overstates the current guarantee and reads stronger than the
  repo's actual declarative model.

  I would recommend aligning this doc to `academic-pipeline/SKILL.md`,
  `pipeline_state_machine.md`, `ai_research_failure_modes.md`, `academic-paper/SKILL.md`, and
  `shared/ground_truth_isolation_pattern.md` before merge.

  Inline Comment 1
  位置建議：docs/ARCHITECTURE.md:18

  This currently conflicts with the pipeline's canonical checkpoint semantics.

  The repo does not describe only 8 human-gated points. `academic-pipeline/SKILL.md` says each
  stage completion requires user confirmation before proceeding, and
  `pipeline_state_machine.md` explicitly shows a confirmation checkpoint after Stage 2.5 PASS
  and Stage 4.5 PASS as well.

  Because of that, "these are the only places where progress is human-gated" is too strong and
  currently incorrect. I think this section should either:
  - count all user-confirmation checkpoints consistently, or
  - explicitly distinguish between "decision-heavy checkpoints" and "post-stage confirmation
  checkpoints" without implying that 2.5 / 4.5 are confirmation-free.

  Inline Comment 2
  位置建議：docs/ARCHITECTURE.md:83

  There are a few factual mismatches in this Stage 2.5 row against the current repo state.

  First, the 7-mode labels do not match `academic-pipeline/references/
  ai_research_failure_modes.md`. The current taxonomy is:
  1. implementation bug passing AI self-review
  2. hallucinated citation
  3. hallucinated experimental result
  4. shortcut reliance
  5. implementation bug reframed as novel insight
  6. methodology fabrication
  7. frame-lock at early pipeline stage

  Second, "Claim ↔ Reference Map (100% verified)" overstates the current protocol.
  `claim_verification_protocol.md` says pre-review mode samples 30% of claims (minimum 10),
  while 100% claim verification is the final-check mode.

  I would revise this row to match the current integrity protocol verbatim, otherwise this doc
  will become a stronger contract than the implementation.

  Inline Comment 3
  位置建議：docs/ARCHITECTURE.md:82

  This row currently compresses `academic-paper` into a 10-agent view, but the canonical skill
  doc still presents it as a 12-agent pipeline.

  In particular, `academic-paper/SKILL.md` includes `citation_compliance_agent` and
  `peer_reviewer_agent` in the main agent team and orchestration flow. By listing only 10
  agents here, this architecture doc changes the apparent responsibility boundaries of Stage
  2.

  If the goal here is to show only the Stage-2-relevant agents, I would say that explicitly
  (for example, "core Stage 2 path") instead of presenting it as the full stage architecture.

  Inline Comment 4
  位置建議：docs/ARCHITECTURE.md:35

  The flow diagram is missing the currently documented coaching transitions between review and
  revision.

  Right now the repo still documents:
  - Stage 3 -> 4 Revision Coaching
  - Stage 3' -> 4' Residual Coaching

  See `academic-pipeline/references/two_stage_review_protocol.md` and `academic-paper-
  reviewer/SKILL.md`.

  Even if you want to keep the diagram high-level, omitting these transitions is incomplete
  for a document that presents itself as the full pipeline architecture. I would either add
  them to the diagram or explicitly state that coaching sub-stages are intentionally elided
  from this view.

  Inline Comment 5
  位置建議：docs/ARCHITECTURE.md:122

  This overstates the current isolation guarantee.

  `shared/ground_truth_isolation_pattern.md` is more careful: it describes a declarative
  architecture, not a runtime-enforced permission system, and it explicitly allows the
  reviewer side to hold a rubric privately as long as that rubric does not leak into the
  candidate-generation context.

  So "No skill ever reads ground-truth / rubrics / gold labels within the same process as
  producing a response" reads stronger than the repo's actual documented contract.

  I would soften this to match the current wording in the pattern doc:
  - `data_access_level` is declarative, not runtime enforcement
  - the key guarantee is that rubric / gold material must not be present in the candidate-
  generation context
  - calibration gold sets are runtime-supplied, not bundled into normal skill loading

Imbad0202:
Thanks — all 5 points addressed in df230d2. Summary of how each was resolved:

**1. Checkpoint model (Inline 1)** — restored every-stage user confirmation per `academic-pipeline/SKILL.md` and `pipeline_state_machine.md`. Dropped the "only 8 human-gated points" framing. The new §1 distinguishes two explicit classes:
- **🧑 Decision-heavy checkpoints** — user chooses a branch or approves a material decision (10 points, now including the two Coaching sub-stages)
- **✓ Post-stage confirmation checkpoints** — machine verification runs first, user acks the report (2.5 and 4.5)

Both classes are user-gated. The previous "machine only" label at 2.5/4.5 is gone.

**2. Stage 2.5 taxonomy + claim % (Inline 2)** — replaced the ad-hoc 7-mode labels with the canonical ones from `ai_research_failure_modes.md`: M1 impl bug / M2 hallucinated citation / M3 hallucinated experimental result / M4 shortcut / M5 reframed bug / M6 methodology fabrication / M7 frame-lock. Replaced "100% verified" with the actual pre-review sampling mode (30%, min 10) per `claim_verification_protocol.md`, and noted Stage 4.5 runs the 100% final-check mode.

**3. academic-paper agent count (Inline 3)** — restored 12 agents. `citation_compliance_agent` and `peer_reviewer_agent` are back. My earlier "these are embedded protocols" rationalization was wrong — they have their own agent files and are in the canonical Phase flow.

**4. Flow diagram coaching transitions (Inline 4)** — added explicit nodes for Stage 3 → 4 Revision Coaching (max 8 rounds) and Stage 3' → 4' Residual Coaching (max 5 rounds) per `two_stage_review_protocol.md`. New `coaching` classDef (green) so they read as distinct sub-stages where the user may engage or skip.

**5. Data isolation section (Inline 5)** — softened to match `ground_truth_isolation_pattern.md` §5 directly: `data_access_level` is declarative not runtime-enforced; reviewer may hold rubric privately; the real guarantee is that rubric content must not leak into the candidate-generating agent's context; calibration gold sets are runtime-supplied.

Also updated the §2 flow diagram legend and §6 Quality Gates table to reflect the three gate classes (🧑 decision-heavy / ✓ integrity-ack / 🤖 CI lint).

Imbad0202:
Inline Comment 1
  位置建議：docs/ARCHITECTURE.md:100（也可順帶指到 :103）

  This is much closer overall, but Stage 3 and Stage 3' still look partially conflated here.

  Right now this Stage 3 row includes `R&R Traceability Matrix`, while the canonical docs
  place that artifact in Stage 3' re-review, not the first-round review. The current repo
  contract is:

  - Stage 3 (full review): `5 review reports + Editorial Decision + Revision Roadmap`
  - Stage 3' (re-review): revision response checklist / new issues / new decision + R&R
  Traceability Matrix

  Relatedly, the Stage 3' row below currently reads as if the same 3 peer reviewers + DA are
  active in re-review, but `academic-paper-reviewer/SKILL.md` currently defines re-review as
  `field_analyst + eic + editorial_synthesizer`.

  I’d recommend tightening these two rows so Stage 3 = first-round review package, and Stage
  3' = verification / traceability package.

  Inline Comment 2
  位置建議：docs/ARCHITECTURE.md:106（也可順帶指到 :196）

  This section still reads slightly stronger than the repo’s current `repro_lock` contract.

  The repo currently allows `repro_lock: null` as an explicit honest opt-out, and `shared/
  artifact_reproducibility_pattern.md` also says Stage 2.5 / 4.5 do not read `repro_lock` at
  runtime. In addition, `check_repro_lock.py` is documented as a standalone validator, not
  something wired into the CI lint suite by default.

  Because of that, phrases like "`repro_lock` populated" and "`invalid hash_timing` -> CI
  errors" overstate the current enforcement model a bit. I think this should be softened to
  reflect the actual contract:
  - key required from v3.3.5+
  - `null` allowed as explicit opt-out
  - populated block validated by `check_repro_lock.py`
  - not currently a Stage 4.5 runtime gate or default CI-enforced behavior

Imbad0202:
Thanks — both points addressed in 253e394.

**Inline 1: Stage 3 vs Stage 3' boundary** — tightened per `academic-paper-reviewer/SKILL.md`:
- **Stage 3 (first-round)**: artifact list is now the canonical *5 review reports (EIC + R1 methodology + R2 domain + R3 interdisciplinary + Devil's Advocate) + Editorial Decision + Revision Roadmap*. **R&R Traceability Matrix removed from this row** — it does not belong here.
- **Stage 3' (re-review)**: agent list tightened to the **narrow re-review team** defined in SKILL.md:181 — *field_analyst + eic + editorial_synthesizer* (3 agents, not the full Stage 3 panel). **R&R Traceability Matrix (Schema 11)** moved here as the canonical artifact, with the "Author's Claim + Verified? columns" spec from the re-review mode doc.

**Inline 2: repro_lock language** — softened to match the actual v3.3.5 contract per `artifact_reproducibility_pattern.md`:
- Stage 4.5 row: changed "`repro_lock` populated" to "`repro_lock` **declared — populated or explicit `null`** (honest opt-out)". Added the explicit note: *"`repro_lock` is not read by the integrity gate at runtime; if populated, `stochasticity_declaration` must be verbatim and is validated by the standalone `check_repro_lock.py` — this is post-hoc documentation, not a runtime block."*
- §6 Quality Gates row: re-labeled as **standalone lint** rather than Stage 4.5 gate; Stage column changed from `4.5` to `—`; failure handling changed from "Lint fails; CI errors" to "Run `check_repro_lock.py <passport>` on demand". Explicitly noted it is **not** CI-wired by default.

The doc now matches the pattern doc's "post-hoc documentation, not runtime validation" framing exactly.

Imbad0202:
Re-checked `253e394`. The Stage 3 / 3' boundary and `repro_lock` wording now align with the
  canonical docs. No further blocking comments from my side.

PR #159, opened 2026-05-19
feat(skills): optional merge of 5 polish/validation skills from academic-writing-skills (MIT)

jiayou20021120-afk:
## Summary

This PR proposes adding **5 post-writing polish & validation skills** from [bahayonghang/academic-writing-skills](https://github.com/bahayonghang/academic-writing-skills) (MIT) to the ARS `skills/` directory, complementing the existing 4 research/writing pipeline skills.

**This is a personal merge from my fork — feel free to close if it does not align with the project's direction.** The five MIT-licensed skills are also installable standalone from the upstream AWS repo; no harm done if ARS prefers to stay focused on the research-to-publication pipeline.

## What this adds

| Skill | Input | Purpose |
|---|---|---|
| `latex-paper-en` | `.tex` | English LaTeX manuscript polish (ChkTeX, latexmk, AXES paragraph coherence, CrossRef + Semantic Scholar bib verification) |
| `latex-thesis-zh` | `.tex` | Chinese degree thesis polish (GB/T 7714-2015, thuthesis/pkuthss/ustcthesis/fduthesis auto-detection) |
| `typst-paper` | `.typ` | Bilingual Typst paper polish with millisecond compile + Hayagriva/BibTeX checks |
| `bib-search-citation` | `.bib` | Local BibTeX/BibLaTeX library search (Zotero-export compatible) |
| `aiwei-zh` | `.md/.txt/.tex/.typ/.docx` | Chinese AI-tone detection and removal (Huangfu Boyuan 2026 six-dimension framework + em-dash hard gate) |

## Why these five (and not the other two from AWS)

Two AWS skills were **deliberately excluded** to avoid routing collisions with existing ARS skills:

- `paper-audit` substantially overlaps with `academic-paper-reviewer` (both run multi-reviewer simulation + synthesis)
- `industrial-ai-research` is a vertical specialization of `deep-research`

## Routing positioning

The five new skills target the **post-writing layer** ARS currently does not cover:
- `academic-paper` produces a draft *from scratch* → `latex-paper-en` / `latex-thesis-zh` / `typst-paper` polish a draft you *already have*
- `aiwei-zh` is a Chinese-specific hard-gate complement to `academic-paper`'s general-purpose `writing_quality_check.md` (em-dash / quote typographic gates that the general check does not run)
- `bib-search-citation` complements `academic-paper`'s `citation-check` mode by searching a local `.bib` library *before* citing

## License compatibility

- ARS core remains **CC-BY-NC-4.0** (unchanged)
[...]

PR #187, opened 2026-05-21
docs(readme): add Academic Writing Skills as a companion plugin (per #159)

jiayou20021120-afk:
## Summary

Follow-up to #159, taking the path you suggested: a small adjacent-projects README pointer instead of a core merge.

Adds a `## Companion: Academic Writing Skills (post-writing polish)` section right after the existing `## Companion: Experiment Agent`, mirroring that pattern. It points to [bahayonghang/academic-writing-skills](https://github.com/bahayonghang/academic-writing-skills) (AWS, MIT) as an independently-installable companion for the post-writing polish layer (LaTeX/Typst compile checks, bibliography hygiene, Chinese AI-tone removal) that sits after ARS Stage 2 (WRITE) / Stage 5 (FINALIZE).

## Why this shape

- **Zero maintenance burden for ARS** — the two suites install independently with no modification to either, so none of the upstream-drift / license-provenance / routing-contract concerns from #159 apply.
- **Consistent with existing docs** — same length and structure as the experiment-agent companion block.
- **One file, +18 lines** — `README.md` only. No skills, no manifest changes, no NOTICE.

## Changes

- `README.md` — one new `## Companion` subsection (18 lines added, nothing removed)

## Test plan

- [ ] Companion section renders correctly on GitHub
- [ ] Both companion blocks (experiment-agent + AWS) read consistently
- [ ] No other files touched (`git diff --stat` shows only `README.md`)

Thanks again for the thoughtful close on #159 — this is the cleaner shape.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,24 @@ ARS Stage 2 WRITE     →  write paper with verified experiment results
 
 ---
 
+## Companion: Academic Writing Skills (post-writing polish)
+
+If your manuscript is already drafted and needs format-level polish (LaTeX/Typst compile diagnosis, bibliography hygiene, or AI-tone removal), [Academic Writing Skills](https://github.com/bahayonghang/academic-writing-skills) (AWS, MIT) covers the post-writing layer after ARS Stage 2 (WRITE) / Stage 5 (FINALIZE).
+
+```
+ARS Stage 2 WRITE / Stage 5 FINALIZE  →  draft manuscript (.tex / .typ)
+        ↓
+  academic-writing-skills              →  compile checks · bib verification · de-AI polish
+        ↓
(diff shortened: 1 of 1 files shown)


Imbad0202:
Thanks @jiayou20021120-afk for following through on this, and for the clean diff.

After more thought, I'd like to walk back the companion-pointer invitation from #159. On reflection, the main README isn't the right home for third-party pointers, and that's a call I should have made before suggesting it. The reason is on my side, not on the PR.

AWS stands well on its own, and users who want to combine the two suites can already do so independently. If linking is useful in the other direction, the AWS README would be a natural home.

Sorry for the about-face, and thanks again for the careful work on both #159 and this PR.

PR #237, opened 2026-05-23
Add optional arxiv_id to literature_corpus schema and extend validation coverage

### LINKED ISSUE #234: #182 implementation: where does arXiv ID live for verify_citation input?
## Context

`#182` `verification_gate.verify_citation(citation_key, manifest, corpus, cache) -> CitationVerificationOutcome` per the spec at `docs/design/2026-05-21-v3.10-182-promote-citation-gate-spec.md` Delta 5 reads the citation `corpus` (passport-shaped, per `shared/contracts/passport/literature_corpus_entry.schema.json`) and dispatches to the four resolvers (Crossref / OpenAlex / Semantic Scholar / arXiv).

The new `scripts/arxiv_client.py` (per #182 Delta 1) needs an arXiv ID to perform the DOI-equivalent lookup. But `literature_corpus_entry.schema.json` does **not** currently declare an `arxiv_id` field, and #182 spec Delta 1 also does not add one — the only Delta 1 schema change is the `contamination_signals.arxiv_unmatched` boolean.

This leaves a gap: where does `verify_citation` actually source the arXiv ID from when running the arXiv resolver?

## Three candidate paths

**Path A — Parse `source_pointer` URI.** The existing `source_pointer` field already accepts arXiv URLs like `https://arxiv.org/abs/2401.12345` per the schema description. `arxiv_client` extracts the ID from this URI when applicable.

- Pro: no schema change; reuses an existing field
- Pro: backward-compatible with all existing corpora
- Con: parsing logic lives in `arxiv_client` rather than as a structured schema field
- Con: schema does not currently enforce that `source_pointer` for arXiv citations carries the canonical arXiv URL form; some corpora may carry the PDF URL or the abstract URL with different patterns
- Con: old-style arXiv IDs (`hep-th/9711200` per #182 OQ-2) need their own parse logic distinct from new-style (`2605.07723`)

**Path B — Parse `doi` field.** arXiv issues DOIs of the form `10.48550/arXiv.<id>` for papers that opt in. `arxiv_client` could detect this DOI pattern and extract the ID.

- Pro: no schema change
- Con: only works for papers that have an arXiv DOI; many older arXiv papers don't
- Con: contradicts #184 §3.1.1 wording \"valid arXiv ID (no DOI)\" — implies the design intent is that arXiv-only citations have NO `doi` field at all
- Con: parsing logic is even more fragile than Path A

**Path C — Add `arxiv_id` field to `literature_corpus_entry.schema.json`.** Add an optional `arxiv_id` field of pattern `^([a-z\-]+/\d{7}|\d{4}\.\d{4,5}(v\d+)?)$` (accepting both old-style and new-style per #182 OQ-2).

- Pro: structured, queryable, validatable
- Pro: adapter contracts (`shared/contracts/passport/` adapter docs) can explicitly declare arXiv ID extraction is a first-class adapter responsibility
- Pro: tests can assert `arxiv_id` presence/absence per `kind`
- Con: schema change with migration concern (existing corpora won't have it, but it's optional so no break)
- Con: needs a small migration helper for users who want to backfill `arxiv_id` on existing v3.7.3+ corpora

## #184 Phase 1a interim decision (do not block this issue)

The #184 Phase 1a gold-set implementation carries `arxiv_id` as a **tuple-level field** (outside `corpus_entry`) for 10 `valid_arxiv` tuples. This is a forward-looking decision: when this issue resolves, Phase 1b harness logic translates the tuple-level field to whatever shape #182 lands on. No #184 work needs to wait for this issue.

## Recommended direction (open to discussion)

Path C looks cleanest:
- It makes the schema honest about a piece of metadata `arxiv_client` already needs
- It avoids parse-logic-in-resolver coupling
- The migration cost is small (optional field, no break to existing corpora)
- It composes with the existing v3.7.3 `obtained_via` / v3.9.0 contamination-signal additive pattern

But the call is genuinely open — Path A could be reasonable if the team prefers minimal schema surface.

## What this issue is NOT

- Not a request to refactor `arxiv_client` after it ships — this needs to be resolved **during** #182 implementation before `arxiv_client` is wired
- Not a #184 blocker (Phase 1a tuple-level `arxiv_id` works regardless of the resolution)
- Not a v3.10 ship blocker for #184; it IS a `#182` implementation-phase decision

## References

- Spec #182: `docs/design/2026-05-21-v3.10-182-promote-citation-gate-spec.md` Delta 1 + Delta 5
- Spec #184: `docs/design/2026-05-21-v3.10-184-extend-eval-harness-spec.md` §3.1.1 (gold-set arXiv tuple expected shape)
- Schema: `shared/contracts/passport/literature_corpus_entry.schema.json`
- Umbrella: #198

madtriceps:
### Description
This PR introduces support for `arxiv_id` in the `literature_corpus` schema and extends validation coverage to ensure robustness across different ID formats and CLI inputs.

#### Key Changes
* **Schema Update:** Added optional `arxiv_id` field to `literature_corpus_entry.schema.json`.
    * Supports new-style IDs (e.g., `2401.12345`).
    * Supports legacy-style IDs (e.g., `hep-th/9711200`).
* **Schema Validation:** Implemented comprehensive testing for:
    * Valid new-style arXiv IDs.
    * Valid legacy arXiv IDs.
    * Invalid/malformed ID formats.
* **Integration Testing:** Extended `ars-mark-read` coverage:
    * Added tests using actual passport fixtures.
    * Verified handling of citation keys containing spaces.

#### Objective
This change keeps the focus on improving schema validation and data integrity while ensuring that CLI integration remains stable and compliant with the updated corpus structure.

---
Closes #234

### DIFF
--- a/academic-pipeline/references/adapters/overview.md
+++ b/academic-pipeline/references/adapters/overview.md
@@ -43,6 +43,7 @@ Refer to the [`literature_corpus_entry` schema](../../../shared/contracts/passpo
 | `abstract` | string | PRIVATE FIELD. |
 | `adapter_name` | string | Optional. |
 | `adapter_version` | string | — |
+| `arxiv_id` | string | Optional arXiv identifier for arXiv-hosted works. |
 | `contamination_signals` | object | v3.7.3 + v3.9.0 contaminated-source advisory field (spec v3.7.3 §3.2 + v3.9.0 §3.4–§3.5). |
 | `contamination_signals_backfilled_at` | string | v3.7.3 backfill provenance (issue #105). |
 | `description_last_audit` | null \| string | v3.7.1 trust-chain field. |
--- a/scripts/adapters/tests/test_literature_corpus_entry_schema.py
+++ b/scripts/adapters/tests/test_literature_corpus_entry_schema.py
@@ -235,6 +235,47 @@ def test_valid_obtained_at_format_passes():
     _validator(schema).validate(entry)
 
 
+def test_valid_arxiv_id_passes():
+    schema = _load_schema()
+    entry = {
+        "citation_key": "chen2024",
+        "title": "T",
+        "authors": [{"family": "C"}],
+        "year": 2024,
+        "source_pointer": "https://arxiv.org/abs/2401.12345",
+        "arxiv_id": "2401.12345",
+    }
+    _validator(schema).validate(entry)
+
+
+def test_valid_legacy_arxiv_id_passes():
+    schema = _load_schema()
+    entry = {
+        "citation_key": "niels1997",
+        "title": "Legacy arXiv",
+        "authors": [{"family": "Niels"}],
+        "year": 1997,
+        "source_pointer": "https://arxiv.org/abs/hep-th/9711200",
+        "arxiv_id": "hep-th/9711200",
+    }
+    _validator(schema).validate(entry)
+
+
+def test_invalid_arxiv_id_rejected():
+    from jsonschema.exceptions import ValidationError
+    schema = _load_schema()
+    entry = {
+        "citation_key": "invalid2024",
+        "title": "Bad arXiv",
+        "authors": [{"family": "C"}],
+        "year": 2024,
+        "source_pointer": "https://arxiv.org/abs/2401.12345",
+        "arxiv_id": "arXiv:2401.12345",
+    }
+    with pytest.raises(ValidationError):
+        _validator(schema).validate(entry)
+
+
 # --- v3.7.3 contamination_signals (L3-2) -------------------------------
 # Motivation: Zhao et al. arXiv:2605.07723 (2026-05).
 
--- a/scripts/adapters/tests/test_literature_corpus_entry_schema.py
+++ b/scripts/adapters/tests/test_literature_corpus_entry_schema.py
@@ -276,6 +276,57 @@ def test_invalid_arxiv_id_rejected():
         _validator(schema).validate(entry)
 
 
+def test_valid_legacy_arxiv_id_with_version_passes():
+    schema = _load_schema()
+    entry = {
+        "citation_key": "hepth1997v2",
+        "title": "Legacy arXiv versioned",
+        "authors": [{"family": "C"}],
+        "year": 1997,
+        "source_pointer": "https://arxiv.org/abs/hep-th/9711200v2",
+        "arxiv_id": "hep-th/9711200v2",
+    }
+    _validator(schema).validate(entry)
+
+
+def test_valid_legacy_arxiv_id_with_subject_class_passes():
+    schema = _load_schema()
+    entry = {
+        "citation_key": "math0703087",
+        "title": "Subject-class arXiv",
+        "authors": [{"family": "C"}],
+        "year": 2007,
+        "source_pointer": "https://arxiv.org/abs/math.AG/0703087",
+        "arxiv_id": "math.AG/0703087",
+    }
+    _validator(schema).validate(entry)
+
+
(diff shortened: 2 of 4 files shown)


Imbad0202:
Thanks for picking this up — and for the run of PRs over the last few days.

Two things before merge:

**arxiv_id regex doesn't cover legacy IDs.** Tested with `re.match()`:

```
hep-th/9711200v2             fail   (legacy + version)
math.AG/0703087              fail   (subject-class)
cond-mat.mes-hall/0701435    fail
cs.CL/0605041                fail
q-bio.BM/0401004v1           fail
math/0611800v2               fail
```

Pattern:

```
^([a-z][a-z0-9-]*(\.[A-Za-z0-9-]+)?/\d{7}(v[1-9]\d*)?|\d{4}\.\d{4,5}(v[1-9]\d*)?)$
```

Boundary tests to add: legacy versioned, dotted subject-class, 4-digit new-style (`0704.0001`), omitted `arxiv_id` (backward-compat).

Refs: [arXiv identifier scheme](https://info.arxiv.org/help/arxiv_identifier.html), [for services](https://info.arxiv.org/help/arxiv_identifier_for_services.html).

**`tests/test_mark_read_args.py` changes are unrelated and use schema-invalid keys.** `"wang 2023 formative"` violates `citation_key.pattern: ^[A-Za-z][A-Za-z0-9_:-]*$`. Two options:

- swap to a schema-valid special-char key (`wang_2023`, `wang:2023`)
- split into its own PR (the CLI permissive vs schema strict question is real and deserves its own thread)

madtriceps:
> Thanks for picking this up — and for the run of PRs over the last few days.
> 
> Two things before merge:
> 
> **arxiv_id regex doesn't cover legacy IDs.** Tested with `re.match()`:
> 
> ```
> hep-th/9711200v2             fail   (legacy + version)
> math.AG/0703087              fail   (subject-class)
> cond-mat.mes-hall/0701435    fail
> cs.CL/0605041                fail
> q-bio.BM/0401004v1           fail
> math/0611800v2               fail
> ```
> 
> Pattern:
> 
> ```
> ^([a-z][a-z0-9-]*(\.[A-Za-z0-9-]+)?/\d{7}(v[1-9]\d*)?|\d{4}\.\d{4,5}(v[1-9]\d*)?)$
> ```
> 
> Boundary tests to add: legacy versioned, dotted subject-class, 4-digit new-style (`0704.0001`), omitted `arxiv_id` (backward-compat).
> 
> Refs: [arXiv identifier scheme](https://info.arxiv.org/help/arxiv_identifier.html), [for services](https://info.arxiv.org/help/arxiv_identifier_for_services.html).
> 
> **`tests/test_mark_read_args.py` changes are unrelated and use schema-invalid keys.** `"wang 2023 formative"` violates `citation_key.pattern: ^[A-Za-z][A-Za-z0-9_:-]*$`. Two options:
> 
> * swap to a schema-valid special-char key (`wang_2023`, `wang:2023`)
> * split into its own PR (the CLI permissive vs schema strict question is real and deserves its own thread)

Understandable, addressed.

Imbad0202:
Verified the regex against all the legacy cases (hep-th/9711200v2, math.AG/0703087, cond-mat.mes-hall/0701435, cs.CL/0605041, q-bio.BM/0401004v1, math/0611800v2) and the new-style ones (2401.12345, 2401.12345v2, 0704.0001) — all pass, malformed inputs (`arXiv:` prefix, empty, wrong digit count, v0) all reject. Boundary tests cover the cases well. Merging now. Thanks for sticking with this through two rounds.

PR #240, opened 2026-05-23
docs: add clinical epistemic status example

### LINKED ISSUE #183: Add mandatory epistemic_status field with silent-upgrade detection
## Summary

Add a mandatory `epistemic_status` field to every claim-bearing artifact produced by the pipeline (drafts, summaries, fact-checks, systematic-review outputs). Block silent upgrades from weaker to stronger statuses without a verification trail.

## Why

A recent Nature paper on automating scientific discovery reached novel-looking in-vitro findings but the authors explicitly stated: "would of course require validation in a suitable disease model and ultimately in a randomized, placebo-controlled trial." (Ghareeb et al., 2026.) The paper itself never crossed from hypothesis into validated conclusion — and was careful to mark that line.

Academic literature pipelines have the same risk in a milder form: an LLM can summarize a paper's *speculative discussion* using the same prose register it uses for *validated findings*, and the reader cannot tell which is which. This is not a hallucination problem — the citation can be perfectly correct — it is an epistemic-status problem.

## Proposal

Require every claim-bearing block to carry an `epistemic_status` enum from a small fixed vocabulary:

- `hypothesis` — proposed but unvalidated
- `association` — statistical association observed
- `in_vitro_evidence` — bench evidence only
- `observational_evidence` — observational study, not causal
- `causal_claim` — causal inference from an intervention design
- `validated_conclusion` — replicated, gold-standard evidence

Rules:

1. Status is **assigned by an extractor that reads the source**, not by the summarizer. The extractor returns the strongest status it can support with a source-text quote.
2. Downstream summaries **cannot upgrade** the status. If a paper supports `in_vitro_evidence`, the summary cannot phrase it as a `causal_claim`.
3. Outputs render the status visibly (e.g., colored badge in HTML, bracketed tag in markdown: `[in_vitro_evidence]`).
4. A `silent_upgrade_check` runs as part of post-generation review: it compares the status the summarizer's prose register implies (causal language? "shows that X causes Y"?) against the recorded status. Mismatches block "ready for human review."

## Acceptance

- Vocabulary defined in `epistemic_status.md` with examples for each level.
- Extractor module assigns and quotes-source.
- Summarizer prompts updated to forbid status-upgrade language.
- `silent_upgrade_check` lint runs on draft output; CI fails the draft if any mismatch is found.
- Test fixtures cover: (a) summarizer trying to upgrade in_vitro to causal, (b) extractor finding only association but reader trying to cite it as cause, (c) hypothesis paragraph followed by a sentence written in conclusion register.

## Out of scope

- Cross-paper status reconciliation (e.g., meta-analysis combining multiple in_vitro studies into observational evidence). Follow-up.

## Priority

P0. Without this, the pipeline can produce technically-correct citations attached to overstated claims — which is harder to detect and more damaging to credibility than outright hallucination.

---

## Acknowledgment

The motivation framing in this issue draws directly on:

> Ghareeb, A. E., Chang, B., Mitchener, L., Yiu, A., Szostkiewicz, C. J., Shved, D., Gyimesi, G. J., Laurent, J. M., Wright, S. M., Razzak, M. T., White, A. D., Finnemann, S. C., Hinks, M. M., & Rodriques, S. G. (2026). **A multi-agent system for automating scientific discovery.** *Nature.* https://doi.org/10.1038/s41586-026-10652-y

Robin (FutureHouse) — Crow + Falcon (literature search) and Finch (data analysis) specialist agents driving closed-loop hypothesis generation for biology. The paper's empirical findings on citation hallucination rates, deterministic-vs-agentic workflows, source-verification gates, and the boundary between hypothesis and validated conclusion are what made the issue above feel actionable rather than abstract. Credit where it's due.

Code/data: https://github.com/Future-House/robin · https://github.com/Future-House/finch · NCBI SRA PRJNA1464762.

2023Anita:
## Summary

- Add a clinical epistemic-status example for detecting silent upgrades in health research writing.
- Show how an observational clinical source can be overstated into causal or routine-care recommendation language.
- Link the new example from `academic-paper/SKILL.md`.

## Motivation

Related to #183.

[skip-closes-check] This PR is intentionally a docs/example contribution and does not close #183, because #183 still requires a project-level schema/vocabulary decision plus implementation of extractor, lint, and fixtures.

This is intentionally a docs/example contribution, not a schema or vocabulary decision. The example preserves the design question noted in #183: the existing 5-tier claim-confidence axis and the proposed research-design-stage axis should not be collapsed without a project-level decision.

## What the example covers

- Association-to-cause upgrade
- Observational evidence-to-clinical recommendation upgrade
- Hedge dropping under summary/compression pressure
- Population-scope widening
- Synthetic mini fixtures for future tests or examples

## Verification

- `git diff --check`
- `python3 scripts/check_spec_consistency.py`
- `python3 scripts/check_version_consistency.py`
- `python3 scripts/check_data_access_level.py`
- `python3 scripts/check_task_type.py`

## Risk

Docs/example-only. No agent behavior, schema, CI, or runtime code changes.

### DIFF
--- a/academic-paper/SKILL.md
+++ b/academic-paper/SKILL.md
@@ -360,7 +360,7 @@ See `agents/intake_agent.md` for the complete field definitions of the Phase 0 c
 
 **Templates** (11 files in `templates/`): `imrad`, `literature_review`, `case_study`, `theoretical_paper`, `policy_brief`, `conference_paper`, `latex_article_template.tex`, `bilingual_abstract`, `credit_statement`, `funding_statement`, `revision_tracking` (4 status types).
 
-**Examples** (7 files in `examples/`): `imrad_hei_example`, `literature_review_example`, `plan_mode_guided_writing`, `chinese_paper_example`, `revision_mode_example`, `revision_recovery_example`, `clinical_citation_verification_checklist`.
+**Examples** (8 files in `examples/`): `imrad_hei_example`, `literature_review_example`, `plan_mode_guided_writing`, `chinese_paper_example`, `revision_mode_example`, `revision_recovery_example`, `clinical_citation_verification_checklist`, `clinical_epistemic_status_example`.
 
 ---
 
--- a/academic-paper/examples/clinical_epistemic_status_example.md
+++ b/academic-paper/examples/clinical_epistemic_status_example.md
@@ -1,0 +1,95 @@
+# Clinical Epistemic Status Example
+
+This example shows how to detect silent epistemic upgrades in clinical and health research writing. It is a documentation example, not a schema decision for the final `epistemic_status` vocabulary.
+
+Use this when a draft turns preliminary, observational, in-vitro, or hypothesis-level evidence into stronger clinical or causal language.
+
+---
+
+## Scenario
+
+A draft contains the following sentence:
+
+> GLP-1 receptor agonist use prevented cardiovascular hospitalization in patients with type 2 diabetes and should be prioritized in routine care.
+
+The cited source is a retrospective electronic health record cohort study. The source text says:
+
+> In this retrospective cohort, GLP-1 receptor agonist exposure was associated with a lower rate of cardiovascular hospitalization after adjustment for measured covariates. Residual confounding remains possible, and prospective trials are needed before practice recommendations can be made.
+
+The citation may be real and correctly formatted, but the draft still overstates the source.
+
+---
+
+## Status Mapping
+
+This example preserves both epistemic axes mentioned in issue #183:
+
+| Axis | Conservative label | Why |
+|---|---|---|
+| Current 5-tier claim-confidence axis | `Preliminary` or `Supported` | The finding has empirical support, but it is observational and not necessarily replicated. |
+| Proposed research-design axis | `observational_evidence` | The study design supports association language, not an intervention-level causal claim. |
+| Disallowed upgrade | `causal_claim` / `validated_conclusion` | The source explicitly says residual confounding remains possible and trials are needed. |
+
+Do not collapse these axes without a project-level vocabulary decision. For this example, the safety rule is simpler: the draft cannot use stronger language than the source supports.
+
+---
+
+## Silent Upgrade Audit
+
+| Source signal | Draft language | Upgrade type | Safer rewrite |
+|---|---|---|---|
+| "retrospective cohort" | "prevented" | Observational to causal | "was associated with a lower rate of..." |
+| "associated with" | "prevented" | Association to cause | "was associated with..." |
+| "after adjustment for measured covariates" | omitted | Confounding hedge dropped | "after adjustment for measured covariates, with residual confounding possible..." |
+| "prospective trials are needed" | "should be prioritized in routine care" | Evidence summary to clinical recommendation | "does not establish routine-care prioritization without prospective trial evidence..." |
+| "patients with type 2 diabetes" | broad clinical population implied | Population scope widened | "in the studied type 2 diabetes cohort..." |
+
+---
+
+## Checklist
+
+Before accepting a clinical claim, verify:
+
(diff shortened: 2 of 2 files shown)


Imbad0202:
Thanks for this. The GLP-1 source-vs-overclaim framing works well, and the upgrade-kinds map onto standard EBM vocabulary cleanly.

Four findings below. The two P2s need fixes before this can merge as a #183 reference:

**P2: yaml verdict block** (line 68): `current_5_tier_status: "Preliminary or Supported"` isn't a valid #183 enum (5-tier is closed, single-value), and the block mixes non-schema fields (`upgrade_kind`, `clinical_safety_note`) without disclaimer. Either add an "illustrative, not schema" note above the block, or make the block match the schema with single values.

**P2: "both epistemic axes" framing** (lines 25, 30, 69, 95): #183 spec treats the 5-tier enum as primary/closed; the biomedical labels are optional/future `evidence_type` examples, not a co-equal axis. Re-frame as "illustrative future-evidence-type labels".

**P3: GLP-1 + CV hospitalization too broad** (lines 13, 44): LEADER/SUSTAIN-6/REWIND show MACE benefit for specific GLP-1 RAs; ADA 2026 recommends them for defined T2D risk groups. Narrowing line 44 to "this cited EHR cohort alone" avoids accidentally overclaiming while warning against overclaiming.

**P3: "Mini Fixtures" overstates** (line 83): bottom table is concept sketches, not test fixtures. Rename to "Mini Examples" or fill in expected schema fields per row.

Two paths for the P2s:
- you push a fix and we re-review
- merge as-is, and I'll take a follow-up commit for the disclaimer + axes wording

Whichever fits your workflow.

Imbad0202:
Thanks for the fast turnaround on the four findings. All four are addressed.

A second pass surfaced two more substantive things and a few smaller ones. The two P2s would push this past the bar for a #183 reference example, since readers will quote the YAML block and checklist directly.

**P2: `current_5_tier_status: "Supported"` is attached to the wrong claim** (line 65-68). The yaml block's `claim_under_review` is the *unsafe* draft sentence ("prevented... should be prioritized in routine care based on this study alone"). Marking that claim as "Supported" reads as endorsement of the overclaim itself. Only the source-side association is supported. Two options: rename the field (e.g., `source_finding_status: "Supported"` plus `draft_claim_problem: "silent_upgrade_detected"`), or add an inline comment clarifying that `current_5_tier_status` describes the source finding, not the draft claim.

**P2: yaml `claim_under_review` drops the population scope** (line 65 vs line 13). Line 13's unsafe draft says "in patients with type 2 diabetes"; line 65's yaml omits it. `required_rewrite` (line 75) similarly says "studied retrospective cohort" but doesn't bind it to type 2 diabetes. This weakens the population-scope audit the example is built to demonstrate. Suggest: include "in patients with type 2 diabetes" in `claim_under_review`, and make `required_rewrite` say "in the studied type 2 diabetes EHR cohort" or equivalent.

**P3: `required_rewrite` drops the adjustment hedge** (line 75). The audit table (line 43) flags "after adjustment for measured covariates" as a dropped hedge, but `required_rewrite` doesn't restore it. Add the phrase back to keep the table and the rewrite in sync.

**P3: yaml fence loses the illustrative boundary when copied** (line 64-79). The disclaimer at line 79 only fires for readers who reach the end. If someone copies just the yaml block, they get something that looks like a schema example. Suggest adding `# Illustrative only; not a #183 schema proposal.` as the first line inside the fence.

**P3: checklist missing endpoint / temporal / absolute-vs-relative dimensions** (lines 53-58). A draft could silently swap "CV hospitalization" for "MACE" or "mortality", or convert an HR into an absolute benefit, without tripping any of the 6 current items. One more item covering endpoint / effect-size / timeframe scope would close this.

Same two paths as before:

- you push a fix and I re-review
- merge as-is, and I take a follow-up commit for the two P2s + P3s

Whichever fits.

PR #335, opened 2026-06-05
fix: update post-squash codex review command

### LINKED ISSUE #334: post-squash-review.yml prints a broken codex command (--model/--reasoning not flags; --commit rejects prompt)
## Summary

`post-squash-review.yml` emits a `codex review` invocation that **fails on the installed codex CLI** — both in the runner's own attempt (lines 86-87) and in the "Run locally" command string it prints into every advisory issue (line 127). Discovered while clearing the 10-issue post-ship-review backlog (#254–#319): the printed command pasted as-is errors out on every one of those issues.

## The two breakages

The workflow uses:
```bash
codex review --commit <SHA> --title '...' --model gpt-5.5 --reasoning xhigh
```

On the installed codex CLI this fails for two independent reasons:

1. **`--model` / `--reasoning` are not `codex review` flags.** Model + reasoning are set via config overrides:
   ```
   error: unexpected argument '--model' found
   tip: to pass '--model' as a value, use '-- --model'
   ```
   The correct form is `-c model="gpt-5.5" -c model_reasoning_effort="xhigh"`.

2. **`--commit <SHA>` mode rejects a `[PROMPT]` argument.** When the boundary/focus prompt is supplied alongside `--commit`:
   ```
   error: the argument '[PROMPT]' cannot be used with '--commit <SHA>'
   ```
   So `--commit` review takes no prose prompt — model/reasoning must go through `-c`, and any custom instructions cannot be passed in `--commit` mode.

## Locations

- `.github/workflows/post-squash-review.yml:83-88` — the runner's own `codex review` call (lines 86-87 carry the bad flags). The runner currently has codex unavailable anyway (the issues say "codex CLI not available on runner"), but if codex is ever installed on the runner this call fails as written.
- `.github/workflows/post-squash-review.yml:125-127` — the **"Run locally"** string printed verbatim into each advisory issue. This is the higher-impact one: every triager who copy-pastes it hits the error.

## Working invocation (verified during the #254–#319 sweep)

```bash
codex review \
  --commit <SHA> \
  --title 'post-squash integration review (PR #<N>)' \
  -c 'model="gpt-5.5"' \
  -c 'model_reasoning_effort="xhigh"' \
  --enable web_search_cached
```

(No `[PROMPT]` arg in `--commit` mode. `--enable web_search_cached` lets codex look up docs/APIs during review, matching the `/codex` skill's behavior.)

## Fix

Update both sites to the `-c model=... -c model_reasoning_effort=...` form and drop any prose-prompt argument from the `--commit` path. Optionally pin the codex CLI version in the runner step (or assert a minimum version) so a future flag rename surfaces as an explicit failure rather than a silently-wrong printed command.

## Scope note

CI-infra fix, no product-code impact. Found via the post-ship-review backlog clear; tracking so the next CI touch corrects the template that feeds every future advisory issue.

puneetdixit200:
## Summary

Fixes #334.

This updates `.github/workflows/post-squash-review.yml` so both the runner invocation and the printed local fallback command use the current `codex review` CLI form:

- model and reasoning are passed via `-c` config overrides
- `--commit` mode is kept prompt-free
- `web_search_cached` is enabled explicitly to preserve the intended review behavior

AI assistance was used while preparing this patch. I reviewed the final workflow change and verified the CLI behavior locally.

## Eval impact

No eval impact.

## Checklist

Validation:
- `codex review --commit HEAD --title test --model gpt-5.5 --reasoning xhigh` exits 2 with `unexpected argument --model`
- `codex review --commit HEAD --title test -c 'model="gpt-5.5"' -c 'model_reasoning_effort="xhigh"' --enable web_search_cached --help` exits 0\n- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/post-squash-review.yml"); puts "YAML_OK"'`\n- `git diff --check`

### DIFF
--- a/.github/workflows/post-squash-review.yml
+++ b/.github/workflows/post-squash-review.yml
@@ -83,8 +83,9 @@ jobs:
           codex review \
             --commit "${{ github.sha }}" \
             --title "post-squash integration review (PR #${{ github.event.pull_request.number }})" \
-            --model gpt-5.5 \
-            --reasoning xhigh \
+            -c 'model="gpt-5.5"' \
+            -c 'model_reasoning_effort="xhigh"' \
+            --enable web_search_cached \
             > codex-review.txt 2>&1
           EC=$?
           echo "exit_code=$EC" >> "$GITHUB_OUTPUT"
@@ -124,7 +125,9 @@ jobs:
               echo '```bash'
               echo "codex review --commit ${{ github.sha }} \\"
               echo "  --title 'post-squash integration review (PR #${PR_NUM})' \\"
-              echo "  --model gpt-5.5 --reasoning xhigh"
+              echo "  -c 'model=\"gpt-5.5\"' \\"
+              echo "  -c 'model_reasoning_effort=\"xhigh\"' \\"
+              echo "  --enable web_search_cached"
               echo '```'
             fi
             echo ""


Imbad0202:
Merged, thanks @puneetdixit200. Verified against the installed codex CLI: the old `--model`/`--reasoning` flags error out and the `-c model=...` form is accepted, and the printed "Run locally" command pastes cleanly. Closes #334.

PR #419, opened 2026-06-11
docs: note Windows symlink install caveat

### LINKED ISSUE #413: agents/ relative symlinks break on Windows checkouts and zip installs
Found during an external audit at HEAD 121f904.

The three files in `agents/` (synthesis_agent.md, research_architect_agent.md, report_compiler_agent.md) are relative symlinks into `deep-research/agents/`. On Windows without developer mode / `core.symlinks`, and in zip-download installs, these materialise as one-line text files containing the link path — silently breaking the three plugin agents.

Suggestion: replace with real files generated at release time, or document the constraint.

2023Anita:
## Summary
- Document the Windows / ZIP-download caveat for the three top-level plugin agent symlinks under `agents/`.
- Add a quick local check that shows whether the three entrypoints materialized as symlinks or regular files.
- Point users toward plugin marketplace install, WSL, or Windows symlink-enabled Git checkouts.

## Motivation
On Windows checkouts without symlink support, and in GitHub ZIP downloads, the relative symlinks in `agents/` can materialize as one-line text files containing only the target path. That silently breaks plugin agent discovery. This PR follows the issue's lower-risk option: document the constraint clearly instead of changing the source-of-truth symlink layout.

Closes #413

## Verification
- Ran `git diff --check`.
- Checked that `docs/SETUP.md` includes the three affected agent paths, the `core.symlinks=false` caveat, and the Windows Developer Mode remediation.

## Risk
Docs-only change. No plugin packaging or agent prompt files are changed.

### DIFF
--- a/docs/SETUP.md
+++ b/docs/SETUP.md
@@ -201,6 +201,26 @@ The four skills (`deep-research`, `academic-paper`, `academic-paper-reviewer`, `
 
 **Strongly recommended: open auto-update.** Open the `/plugin` UI, find `academic-research-skills`, and toggle auto-update on. ARS releases roughly every 1–2 weeks; auto-update keeps you in sync without manual refreshes. To refresh manually: `/plugin update academic-research-skills`. (`/plugin marketplace update academic-research-skills` only refreshes the marketplace source list, not the installed plugin itself.)
 
+**Windows / zip-download symlink note:** the plugin surface includes three top-level agent entrypoints under `agents/`:
+
+- `agents/synthesis_agent.md`
+- `agents/research_architect_agent.md`
+- `agents/report_compiler_agent.md`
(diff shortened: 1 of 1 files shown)


Imbad0202:
Thanks for writing this up. The failure mode was real, but #421 (merged earlier today) resolved #413 at the layout level instead: the three `agents/` entrypoints are now materialized as regular files, kept byte-identical to `deep-research/agents/` by a new lint (`scripts/check_agents_mirror_sync.py`). On current main there are no symlinks under `agents/`, so this note would document a failure mode that no longer exists, and the quick check would flag healthy checkouts as broken installs.

Closing on that basis. If you hit any remaining Windows issue on current main, an issue is welcome.

PR #427, opened 2026-06-11
Add teaching-side companion link (Teaching Skills)

YujxZJCN:
As invited in discussion #417 — adds a one-line teaching-side companion reference at the end of the Companion section, matching the experiment-agent slot's framing: what it is, and the SoTL seam where the two suites hand off.

No other changes.


Imbad0202:
Merged — thanks for keeping it exactly to scope. Good luck with the deterministic layer.

Issue #330, opened 2026-06-05
Write-scope guard: authenticated per-invocation grants for draft_writer phase separation


Imbad0202:
## Activation trigger and acceptance (2026-08-08)

The static `phase4_*/** + phase6_*/**` union remains a documented bounded weakness. Activate implementation when the dispatch layer has a concrete authenticated per-invocation grant substrate; do not “fix” it by silently breaking legitimate Phase 6 revision writes.

Acceptance:

- a Phase 4 invocation cannot write Phase 6 paths;
- a Phase 6 invocation cannot write Phase 4 paths unless explicitly granted;
- legitimate Phase 6 patch application remains possible;
- missing, expired, replayed, or tampered grants fail closed;
- main-session and non-Bucket-A semantics stay unchanged;
- tests prove the grant is bound to agent, phase, run, and allowed paths.

This remains separate from #676's instruction/data authority boundary.

---

## Summary

Post-ship review of PR #294 (#134 Slice 1 scoped-write guard, surfaced via #295) found the guard to be **strongly engineered with no live fail-open** on `main`. codex raised 1 P1 + 1 P2: the P1 is an empirically-disproven false-positive (recorded below), the P2 is a **valid refinement of a documented Slice-1 limitation** — filed here at low priority, not a ship-blocking defect.

Independent review: codex gpt-5.5 xhigh + adversarial first-party read + direct invocation of `evaluate_decision`.

---

### Refinement (P2, low priority) — `draft_writer_agent` is the one Bucket A agent given a multi-phase static union

`scripts/ars_phase_scope_manifest.json` classifies `draft_writer_agent` as Bucket A `phase: "4"` but grants it `allowed_write_globs: ["phase4_*/**", "phase6_*/**"]`. Because the PreToolUse payload carries only `agent_type` (not the invocation phase), the guard cannot distinguish a Phase 4 draft invocation from a Phase 6 revision invocation, so a Phase 4 draft call can write Phase 6 dirs — a static-union phase-inflation.

**This is within the documented Slice-1 limitation, not an oversight.** Spec §3.6 (line 123) states explicitly: *"A static union of a multi-phase agent's scopes is weak (it would permit writes to any phase the agent ever touches, defeating the point). Real enforcement... needs per-invocation scope grants — a later slice"* — and Slice 3 (§4 item 3) specifies exactly that (per-invocation grants from the dispatching layer, under a designed authenticated-grant trust boundary).

[...]

Issue #134, opened 2026-05-18
v3.10: Active Conductor Architecture — replace passive routing + autonomous subagent with intent-aware controller + structured task envelopes


Imbad0202:
## Background

#133 surfaced a failure pattern where ARS routing logic silently dispatched a single phase agent (`bibliography_agent`) given ambiguous user input (already-written abstract + pre-collected literature), and that subagent then autonomously executed Phase 3-6 work without independent crosschecks.

Root cause analysis identified **two distinct failure layers**:

- **L1 — Routing failure**: ARS main session auto-routed ambiguous cross-phase materials to a single-phase subagent instead of engaging `pipeline_orchestrator_agent` for reconciliation. No intent-clarification gate triggered.
- **L2 — Subagent scope inflation**: The dispatched subagent treated visible end-goal + write permissions as license to "be helpful" and run the full pipeline.

v3.9.2 (#133 hot-fix) addresses L1 with an intent-clarification gate in SKILL.md routing rules, and L2 with prompt hard fence + optional PreToolUse hook + post-hoc verifier.

This issue tracks the longer-horizon architectural fix: an **active conductor layer** that subsumes both L1 and L2 by making the main session a persistent, intent-aware controller rather than a passive trigger-keyword router.

## Goal

Convert ARS's current passive-routing + autonomous-subagent architecture into an **active conductor / dispatched-executor** architecture where:

1. The main session continuously observes user intent and adapts the pipeline plan in response, rather than only acting on explicit slash commands or trigger keywords.
2. Subagents are dispatched with **structured task envelopes** that explicitly enumerate allowed/forbidden scopes, deliverables, and return contracts — not free-form prose prompts.
3. Subagents return with **structured deliverable reports** that the conductor validates before treating the work as accepted.
4. Cross-phase boundaries are enforced **at dispatch time** (envelope scope) and **at return time** (deliverable validation), not only via runtime PreToolUse hooks.

## Why this is a separate issue (not folded into #133)

- **Scope**: 30-50h engineering vs #133's 12-16h. Touches all 4 SKILL.md files, 30+ agent prompts, introduces 2 new schemas (task envelope + return contract), and reframes the routing/dispatch model.
- **Risk profile**: Architectural change with backwards-compat implications for existing slash commands (`/ars-*`) and direct `Task()` invocations. Needs its own design rounds (codex + Gemini dual-track), not borrowed from #133's design budget.
[...]

PR #450, opened 2026-06-15
Write-scope guard (#134): document draft_writer_agent static-union inflation (#330)

Ritik650:
## Summary

Document the known Slice-1 static-union inflation for `draft_writer_agent`
and clarify that the issue is bounded, intentional, and scheduled to be
addressed through Slice-3 per-invocation grants.

No behavior changes are introduced.

## Eval impact

No eval impact.

## Checklist


Ritik650:
@Imbad0202 This pull request is for "Write-scope guard (#134): draft_writer_agent Bucket A multi-phase static union — refine at Slice 3 (low priority) #330 "

will try to contribute on higher priority issues in future time :)

Imbad0202:
Thanks for this — the Windows path-handling issue you spotted in `_matches_any` is a real one, and the #330 limitation is worth documenting at the manifest surface. I've reworked it into #451 with your authorship credited.

Two adjustments I made there: the `\` → `/` normalization is gated on `os.sep` so it only applies on Windows — applied unconditionally it would rewrite a literal backslash in a POSIX filename into a path separator, which can let a file escape its phase-dir fence (there's a regression test for that now). And the manifest note is worded as a known limitation with the disposition deferred to Slice 3, rather than as a final decision, since #330 is tracking that choice for later.

Closing this in favor of #451. Appreciate the contribution.

PR #453, opened 2026-06-16
feat: add OpenAI-compatible cross-model support (MiMo, DeepSeek)

kzccIneko:
## Summary
- Add `mimo*` and `deepseek*` detection patterns in the cross-model availability check
- Add an OpenAI-compatible Chat Completions API call pattern with `OPENAI_BASE_URL` support
- Add MiMo and DeepSeek to the supported models table and setup guide
- Generic fallback: any model ID with `OPENAI_API_KEY` + `OPENAI_BASE_URL` set routes as `CROSS_MODEL_AVAILABLE=openai_compatible`

## Motivation
Researchers using non-OpenAI providers (Xiaomi MiMo, DeepSeek) with OpenAI-compatible APIs currently cannot use ARS cross-model verification — both the detection case statement and the API endpoint are hardcoded to `api.openai.com`.

This change makes the cross-model verification layer provider-agnostic for any endpoint that speaks the OpenAI Chat Completions protocol, while preserving full backwards compatibility for existing GPT and Gemini users.

## Changes
1. **Detection logic** (`Detecting Available Models`): added `mimo*|deepseek*` → `CROSS_MODEL_AVAILABLE=openai_compatible`. The `*)` catch-all now checks for `OPENAI_API_KEY` + `OPENAI_BASE_URL` before falling back to `none`.
2. **API call pattern**: added a new "OpenAI-Compatible API" section using the standard `/v1/chat/completions` endpoint at `${OPENAI_BASE_URL}`. No grounding guard (web_search not available on compatible providers).
3. **Documentation**: added MiMo and DeepSeek to the Supported Models table and Setup Guide.

## Backwards Compatibility
- Existing `gpt-5.*` and `gemini-*` patterns unchanged
- No new env vars required (uses existing `OPENAI_API_KEY` + `OPENAI_BASE_URL`)
- `OPENAI_BASE_URL` defaults to `https://api.openai.com` when unset

### DIFF
--- a/shared/cross_model_verification.md
+++ b/shared/cross_model_verification.md
@@ -23,12 +23,24 @@ A stress test of 68 AI-generated citations found 31% had problems — and all pa
 
 ## Supported Models
 
+### First-party providers (web-search grounding supported)
+
 | Model | API ID | Provider | Best For |
 |-------|--------|----------|----------|
-| Claude (session model) | _(inherited Claude Code session model — e.g., Fable 5)_ | Anthropic | Primary model (default for all ARS skills) |
+| Claude (session model) | _(inherited Claude Code session model)_ | Anthropic | Primary model (default for all ARS skills) |
 | GPT-5.5 | `gpt-5.5` | OpenAI | Cross-verification — recommended balance (supports `xhigh` reasoning) |
 | GPT-5.5 Pro | `gpt-5.5-pro` | OpenAI | Cross-verification — strongest reasoning (premium pricing: ~6× GPT-5.5) |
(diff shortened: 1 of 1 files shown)


Imbad0202:
Thanks for raising this. Making the cross-model layer work with non-OpenAI providers is useful, since not everyone has an OpenAI account.

We've adopted it in #455 with some changes around the grounding boundary. Compatible providers can't produce the web-search evidence the citation check depends on, so a positive verdict from them is recorded as ungrounded (`NOT_SEARCHED`) instead of counting as agreement. Genuine rejections still pass through as disagreements. We also split out a dedicated `ARS_OPENAI_COMPAT_API_KEY` so a provider key never goes out on `OPENAI_API_KEY`.

Your commits carry the authorship. Closing in favor of #455.

PR #457, opened 2026-06-17
feat: Add GitHub Copilot compatibility files

SID-6921:
﻿## Summary

Add a single, thin GitHub Copilot repository custom instruction file that points users and contributors to authoritative documentation—eliminating duplication and reducing maintenance burden.

## Changes

- Replaced verbose dual-file approach with single .github/copilot-instructions.md
- Removed repetition: now points to existing docs instead of re-describing content
- Added explicit safety caveat to Skip Permissions flag (tool-use confirmation)
- Consolidated documentation pointers into clean reference table

## Why

**Addresses maintainer feedback:**
1. ✅ Eliminates duplication (no content mirrors between Copilot files and READMEs/docs)
2. ✅ Reduces maintenance burden (changes to pipeline only need single update location)
3. ✅ Fixes safety concern (Skip Permissions now has full context warning)
4. ✅ Single thin file (easier to review and maintain)

## Benefits

- Improved IDE experience for Copilot users
- Authoritative docs remain single source of truth
- New contributors quickly oriented to official resources
- Safety warnings prevent accidental security gaps
- No ongoing dual-maintenance cost

## File Changes

- Removed: .instructions.md (duplicate content)
- Removed: redundant .github/.instructions.md
- Updated: .github/copilot-instructions.md (thin pointer-based version)
- Total: ~180 lines (down from ~290)

## Testing

- File follows Copilot instruction format
- All doc links validated
- Safety caveat clearly emphasized
- Discovery path correct (.github/copilot-instructions.md)


Imbad0202:
Thanks for putting this together, and for the thought on Copilot onboarding.

I'm going to pass on this one. The reason is the maintenance burden. These instructions restate a lot of what already lives in the READMEs and `docs/`, so every future change to the pipeline would need to be mirrored here too, and the two files also overlap heavily with each other. For a docs surface I'd want something much thinner that points at the authoritative docs rather than re-describing them.

One thing worth flagging regardless of this PR: the Skip Permissions note here drops the safety caveat that the original docs carry. That flag disables tool-use confirmation, so it shouldn't be presented as a plain recommendation. Worth keeping in mind for any docs work.

Really do appreciate the effort, and leaving this open for now in case you want to react.

Imbad0202:
Thanks for this, @SID-6921 — the pointer-based approach is the right shape for a Copilot instructions file, and I've brought it in.

I rebuilt it locally as #465 so a few corrections could land in one clean commit on top of current `main` (version frontmatter to the current suite, command names, doc-link paths resolving from `.github/`, and the permission-mode guidance to match a recent docs change). Your authorship is preserved via `Co-Authored-By` on that commit, so the credit stays with you.

Closing this in favor of #465. Appreciate the contribution.

Issue #257, opened 2026-05-28
[Kong sub A2] Idea Diversity / Coverage Gap Advisory in Socratic + lit-review modes


Imbad0202:
> Sub-issue of [META] Kong et al. 2026 (#255): https://github.com/Imbad0202/academic-research-skills/issues/255
> Tier: **A** — Tag: see body below

**Tag**: `[ACCEPT-HIGH]` (Claude + Gemini two-model consensus, A2 in convergence report D3)

#### Anchor

- **Paper** §7.4.2 (`kong2026_full.txt:L2148-L2161`, PDF p.38-39): "Diversity is a persistent concern: LLM-generated ideas may cluster in narrow regions of the idea space, limiting their ability to explore genuinely distinct research directions [79]."
- **ARS files**:
  - `deep-research/agents/socratic_mentor_agent.md`
  - `academic-paper/agents/socratic_mentor_agent.md`
  - **`deep-research/agents/bibliography_agent.md:144-156`** — existing `uncovered_topics` / `search-fills-gap` / "known coverage gap" surfacing (gap-detection base; this sub-issue ADDS distributional skew metrics on top)
  - **`academic-paper/agents/literature_strategist_agent.md:120-133, 200-212`** — existing research-gap categories (methodological / population / temporal / geographical / corpus gap dispatch)
  - Related: **#134** Active Conductor Architecture (intent detection layer)

#### Problem

Two related failure modes, both advisory in nature:

1. **Wording / framing pattern similarity advisory** (in Socratic / RQ formulation modes): user's proposed RQ wording falls into common LLM-typical phrasing — "exploring the impact of X on Y" / "investigating the relationship between A and B" — without user realizing the framing itself is generic. Current Socratic mentor does not have a pattern-similarity advisory layer. **NOTE**: this is wording/framing-level advisory, **not idea-quality / novelty judgment** (which would be auto-research) — the goal is to surface "this phrasing looks LLM-typical, is that what you meant?" not to rank user's research idea.
2. **Lit review coverage distributional skew** (in lit-review mode): existing `uncovered_topics` / `search-fills-gap` machinery in `bibliography_agent` covers topic-coverage gaps. What it does NOT cover is **distributional skew on other dimensions** — time concentration (all 2023+), geographic (all North America), methodological (all quantitative), venue tier (all top-3 venues). Current lit-review mode reports topic gaps but does not surface distributional skew.

[...]

PR #468, opened 2026-06-23
Add SkillOpt integration harness for the rq_framing_patterns advisory

brycewang-stanford:
## What

Adds `tools/skillopt/` — an integration harness that wires
[Microsoft **SkillOpt**](https://github.com/microsoft/SkillOpt) onto the existing
`evals/gold/rq_framing_patterns` gold set, so the Kong #257 Socratic
wording-pattern advisory can be **trained as a natural-language skill** and gated
on held-out validation (SkillOpt's rollout → reflect → bounded-edit → validation-gate loop).

**Additive tooling only — no `SKILL.md`, agent, mode, or schema is touched.** The
only change outside the new directory is two `.gitignore` lines for the harness's
regenerated runtime artifacts.

## Why

ARS detects AI-typical RQ wording today with a deterministic regex detector
(`scripts/check_rq_framing_patterns.py`, the twenty `WP01`–`WP20` shells). It scores
`balanced_accuracy = 1.000` on the gold set — but the gold set was *authored to match
those regexes*, so the score reflects construction, not generalization. Any **novel**
AI-typical phrasing outside the 20 shells slips through silently.

This harness offers the complementary capability: train a compact **LLM skill** that
makes the same advisory judgment, measured by the **same metric and acceptance gate**
(FNR < 0.30, FPR < 0.20, balanced accuracy ≥ 0.75). It gives ARS a disciplined,
reproducible way to evolve the advisory toward generalization instead of hand-editing
a prompt — built on the gold-set + threshold infrastructure that already exists.

## Layout

| File | Role |
|---|---|
| `ars_scoring.py` | **Pure** (no SkillOpt/network/key): parse `trigger_advisory`, score + aggregate vs gold; mirrors `check_rq_framing_patterns.py`'s metric. |
| `ars_gold_loader.py` | SkillOpt `SplitDataLoader` over `gold_set.json`. |
| `ars_adapter.py` | SkillOpt `EnvAdapter`: single-turn `claude_chat` rollout + scoring + reflect. |
| `run_ars_skillopt.py` | Launcher — registers the adapter into SkillOpt's env registry **without forking SkillOpt**. |
| `configs/rq_framing_patterns.yaml` | Self-contained config (Claude backend, deliberately small budget). |
| `skills/rq_framing_patterns/initial.md` | Seed skill — the trainable starting point. |
| `tests/test_ars_scoring_offline.py` | Zero-API plumbing test over the real gold set. |

## Testing

- **Offline scoring test passes (zero API):**
  `python -m pytest tools/skillopt/tests/test_ars_scoring_offline.py -q` → 9 passed.
- **End-to-end plumbing verified against SkillOpt `v0.1.0` with zero model calls:**
  register adapter → load gold → deterministic 20/8/12 ratio split → build batch →
  stubbed rollout → score/metrics, all green.
- Resolved the benign `scripts` package-name collision between ARS and SkillOpt
  (PEP 420: SkillOpt's regular package wins; verified with the ARS root first on
  `PYTHONPATH`). Implemented `EnvAdapter.reflect` explicitly because `v0.1.0` marks
  it abstract with no body.

## Scope / next steps

Per `CONTRIBUTING.md`, this is additive tooling (like the existing `evals/` harness),
not a `SKILL.md`/agent change — so no prior issue was required. Running an actual
optimization spends model budget and is left as an **opt-in, maintainer-gated
follow-up**: train, validate the lift with `scripts/run_evals.py --task
rq_framing_patterns`, and bring before/after evidence to an issue before proposing any
trained skill for the pipeline.

Happy to adjust the target task, seed skill, or config to match maintainer preference.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,12 @@ docs/superpowers/
 
 # Python bytecode / pytest cache.
 __pycache__/
+.pytest_cache/
+
+# SkillOpt harness runtime artifacts (tools/skillopt/): the ratio-split
+# materialization and any training checkpoints/outputs are regenerated per run.
+/_generated_splits/
+tools/skillopt/runs/
 
 # Private ROADMAP is intentionally gitignored and maintained outside this public repo
 ROADMAP.md
--- a/tools/skillopt/README.md
+++ b/tools/skillopt/README.md
@@ -1,0 +1,119 @@
+# ARS × SkillOpt integration harness
+
+Wire [Microsoft **SkillOpt**](https://github.com/microsoft/SkillOpt) onto ARS's
+existing `rq_framing_patterns` gold set so the **Socratic wording-pattern
+advisory** (Kong #257) can be *trained as a natural-language skill* and gated on
+held-out validation — the same way SkillOpt trains any agent skill: rollout →
+reflect → bounded edit → validation gate → `best_skill.md`.
+
+> **Status: harness only.** This directory ships the integration plumbing and a
+> seed skill. It does **not** ship a trained skill or modify any `SKILL.md`.
+> Running an optimization (which spends model budget) and proposing a trained
+> skill is a separate, maintainer-gated step — see [Scope & next steps](#scope--next-steps).
+
+## Why this exists (honest framing)
+
+ARS already detects AI-typical research-question wording with a **deterministic
+regex detector** — `scripts/check_rq_framing_patterns.py`, twenty hardcoded shells
+`WP01`–`WP20`. It scores a perfect `balanced_accuracy = 1.000` on the gold set…
+*because the gold set was authored to match those regexes*. That precision is real
+but brittle: any **novel** AI-typical phrasing that no `WP` regex anticipated slips
+through silently.
+
+This harness offers the complementary capability: train a compact **LLM skill**
+that makes the same advisory judgment, measured by the **same metric and the same
+acceptance gate** (FNR < 0.30, FPR < 0.20, balanced accuracy ≥ 0.75). The bet is
+that an LLM skill can generalize to wording the regex can't enumerate while keeping
+false positives on domain-native questions low. SkillOpt is the disciplined way to
+evolve that skill instead of hand-editing a prompt and hoping.
+
+## Layout
+
+| File | Role |
+|---|---|
+| `ars_scoring.py` | **Pure**, no SkillOpt/network/key. Parse `trigger_advisory` from a model answer; score + aggregate vs gold (mirrors ARS's own metric). |
+| `ars_gold_loader.py` | SkillOpt `SplitDataLoader` — reads `evals/gold/rq_framing_patterns/gold_set.json`, normalizes, splits. |
+| `ars_adapter.py` | SkillOpt `EnvAdapter` — single-turn rollout (`chat_target`) + scoring + reflect delegation. |
+| `run_ars_skillopt.py` | Launcher — registers the adapter into SkillOpt's env registry, then delegates to `skillopt-train` / `skillopt-eval`. **No fork of SkillOpt.** |
+| `configs/rq_framing_patterns.yaml` | Self-contained SkillOpt config (Claude backend, small budget). |
+| `skills/rq_framing_patterns/initial.md` | Seed skill — the trainable starting point. |
+| `tests/test_ars_scoring_offline.py` | Zero-API plumbing test over the real gold set. |
+
+## Prerequisites
+
+- `pip install skillopt` (lightweight core: `openai`, `numpy`, `pyyaml`, `httpx`, …).
+- The Claude backend (`backend: claude_chat`) drives your **local Claude CLI**
+  (`claude --print`), so it uses your authenticated `claude` binary rather than a
+  separate metered API key. Optimizer = Opus, target = Sonnet by default (ARS's
+  Opus-for-review / Sonnet-for-execution convention; never Haiku).
+
+## Run
+
+**Offline plumbing test — zero API, zero model calls:**
+
+```bash
+python -m pytest tools/skillopt/tests/test_ars_scoring_offline.py -q
+```
+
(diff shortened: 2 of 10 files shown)


Imbad0202:
Thanks for this. The engineering is genuinely clean: pure offline-testable scoring layer, honest README that names its own weakness, good model routing (Opus optimizer / Sonnet target, no Haiku). I read the whole diff and ran the offline test plus the existing checker. Two blockers and one framing issue mean I can't merge it as-is, but they're worth laying out precisely.

## 1. The `scripts` collision isn't benign. It can break ARS itself.

The README calls this "benign, no action needed," but the failure mode is real and it hits the host repo, not just the harness.

ARS's `scripts/` is a namespace package (no `scripts/__init__.py`, confirmed). The README itself states SkillOpt's `scripts/` is a *regular* package. If that holds, then per PEP 420 a regular package wins over a namespace package once both are on the path, so `from scripts import ...` starts resolving to SkillOpt's `scripts`, not ARS's, in any interpreter where SkillOpt is installed.

That matters because ARS leans on its own `scripts` package heavily:

- 158 `from scripts ...` import statements across 59 of the repo's own `.py` files (modules and tests), counted by AST, excluding `tools/skillopt/`.
- CI runs `python -m scripts.run_evals` and `python -m scripts._eval_threshold_gate` directly (`.github/workflows/eval-harness.yml`, lines 98 and 103).

So a user who follows the README's `pip install skillopt` instruction, then runs ARS's own tooling in that same environment, can hit `ImportError` across the eval harness and a large part of the test suite. The "verified, benign" note only holds inside the launcher's narrow `PYTHONPATH=.` invocation; it doesn't cover the much more common "skillopt installed alongside ARS" case. Worth confirming SkillOpt's actual package shape on your end, since the whole "benign" call hinges on it.

## 2. The acceptance gate isn't actually wired into SkillOpt

The PR's core pitch is "gated on the same metric and acceptance gate the regex detector already clears (FNR < 0.30, FPR < 0.20, balanced accuracy ≥ 0.75)." But `compute_metrics()` / `passes_gate` are only exercised in the offline test. The rollout path emits per-item `hard`/`soft`, and SkillOpt's selection/eval gate never calls the aggregate. As written, the ARS gate is a helper function plus a unit test, not the gate that decides whether a trained skill is accepted. So the headline guarantee ("validated against ARS's own gate") isn't true yet.

## 3. The deeper issue: the problem this solves is mostly already handled

The "Why" frames the gap as: the regex detector (`WP01` to `WP20`) scores 1.000 only because the gold set was authored to it, so novel AI-typical phrasing "slips through silently." That's a real concern about *the regex*. But the regex isn't what runs in front of users.

At runtime, the advisory lives in `deep-research/agents/socratic_mentor_agent.md`. The trigger rule there (line 29) is:

> Fire only when the surface wording clearly matches one or more patterns with high confidence. If the match is weak, ambiguous, or depends on interpreting the idea content, stay silent.

That's an LLM applying judgment against a reference table, not a regex doing string matching. The `WP` list is *examples for the model*, not the production matcher. `scripts/check_rq_framing_patterns.py` is an offline calibration harness; it never runs in the user-facing path, and the advisory is explicitly non-blocking (line 63, "Do not block progression").

So the generalization the PR wants (catching phrasing the 20 shells don't enumerate) is largely what the runtime LLM already does. The harness trains a *new* LLM skill to replace a detector that was never the production judge. The thing you'd be improving isn't on the critical path.

## A direction I think is more useful

If the goal is "make the wording advisory catch novel AI-typical framing better," the highest-leverage moves don't need a training loop or a third-party dependency:

1. **Tighten the runtime reference table directly.** Add a short paragraph to `socratic_mentor_agent.md` stating the `WP` list is illustrative, not exhaustive. The real signal is "this phrasing would survive swapping the nouns for any other field's nouns." Add a few deliberately off-list examples. This targets generalization at the exact point where judgment happens, for the cost of a prompt edit. (This touches an agent definition, so per CONTRIBUTING it'd want a design issue first, but it's a much smaller change to discuss.)

2. **Measure the miss rate before building anything to fix it.** The PR correctly notes the gold set can't show generalization because it was authored to the regex. The missing artifact is a held-out set of *real* AI-typical RQs that sit outside the 20 shells, scored against the current runtime judge. If the miss rate is low, there's nothing to fix. If it's high, that held-out set becomes the acceptance test for any change, including, potentially, a future optimization run. That measurement is the prerequisite this harness skips.

3. **If automated prompt optimization is ever warranted,** it should come *after* (2) shows a real gap, and shouldn't ship a v0.1.0 dependency that shadows `scripts`. For a binary judgment over a 40-item set producing one `.md`, hand-editing plus the repo's existing cross-model review is likely enough.

One more thing worth flagging for whoever picks this up: the gold set and any trained skill here are English-only, against a fixed target model. AI-typical phrasing is both language-specific and model-specific and drifts across model versions, so a frozen, trained English skill carries a re-training cost that the prompt-table approach (judgment against examples) doesn't.

Per CONTRIBUTING, a new tool that introduces a third-party training framework and changes how the advisory is evaluated is closer to "new tooling that affects system behavior" than to a drop-in like the `evals/` harness, so the right next step is a design issue scoping the problem (starting with the miss-rate measurement), rather than landing the harness. Happy to keep talking there. Appreciate the care that went into this.

Imbad0202:
Closing this after no response for over two weeks. The blockers from the review comment stand: the `scripts` package collision can break ARS's own eval harness and test suite in a shared environment, and the ARS acceptance gate is not actually wired into SkillOpt's selection loop, so the headline guarantee does not hold yet.

To keep the useful part of this discussion moving, the direction sketched in the review is now tracked in #501: mark the WP table as illustrative at the runtime judge, add a noun-swap judgment cue plus off-list examples, and then measure the runtime miss rate on a held-out set before building anything heavier. The prompt-edit half is landing shortly; the measurement half stays open in #501. Thanks for pushing on generalization, and for the clean engineering in the harness itself.

Issue #518, opened 2026-07-11
Cross-model gate hardening: risk-stratified sampling, blind disagreement checkpoints, capability allowlist, gpt-5.6-sol promotion bakeoff


Imbad0202:
## Motivation

A 2026-07-11 cross-model consult (gpt-5.6-sol, xhigh) on quality-first placement of the external verifier in a tiered Fable/Opus pipeline produced three recommendations worth tracking, plus one routing correctness gap found while reading the current layer.

## Scope

1. **Risk-stratified sampling at integrity gates** (replaces uniform random 30% in `shared/cross_model_verification.md` §Integrity Verification): verify 100% of citations supporting headline conclusions, numerical claims, causal claims, methods, and disputed claims; random-sample only the low-impact remainder; at Stage 4.5 verify all changed/new claims plus a control sample of unchanged ones.
2. **Blind disagreement checks at irreversible checkpoints** (research-design freeze; final editorial accept/revise/reject): both the primary and the cross-model produce INDEPENDENT structured decisions (cross-model never sees the primary's decision first — anchoring prevention, consistent with the existing integrity-sample design), compare, and run a targeted rebuttal only on material divergences. Disagreement is a review trigger, never a vote.
3. **Do NOT build the generic 6th reviewer** (currently "Planned" in the canonical doc): the consult's counterproductive-conditions list (score averaging, role duplication, findings treated as confirmed defects, majority-vote false confidence, synthesizer context burn) matches ARS's existing anti-patterns. Consider replacing the "Planned" section with the checkpoint-audit design above.
4. **Provider routing vs capability allowlist**: today any `gpt-*`/`gemini-*` id is declared grounded-route-available before the API ever rejects it (`gpt-made-up` included). Separate "which provider endpoint" from "is this id known-good", reusing the provisional/validated distinction introduced for `gpt-5.6-sol` (#516 / PR #515).
5. **`gpt-5.6-sol` promotion bakeoff**: define the non-inferiority run (grounded-search completion, citation-mismatch recall, false-disagreement rate, jq-guard shape stability, p95 latency vs `gpt-5.5`) that flips the recommended default; `scripts/cross_model_smoke_test.sh` is the entry gate.

## Notes

- Items 1-2 are prose-layer changes to the canonical protocol + consuming agents; no schema change expected.
- Related: #517 (model tiering, quality-boost/economy directions) — item 2's checkpoints are the same Stage 2.5/4.5 surfaces.

Issue #262, opened 2026-05-28
[Kong sub B3] Structured Cross-Paper Contradiction Output in `synthesis_agent`


Imbad0202:
> Sub-issue of [META] Kong et al. 2026 (#255): https://github.com/Imbad0202/academic-research-skills/issues/255
> Tier: **B** — Tag: see body below

**Tag**: `[ACCEPT-MEDIUM]` (deep-review finding; **`synthesis_agent` already has prose-level contradiction handling**, this sub-issue adds structured schema + candidate-pair scoping + calibration)

#### Anchor

- **Paper** §7.4.2 (`kong2026_full.txt:L2148-L2161`, PDF p.38-39): "In literature review, systems increasingly retrieve and summarize individual papers well, but still struggle with multi-paper relational reasoning, methodological lineage, and cross-paper contradictions."
- **ARS files**:
  - **`deep-research/agents/synthesis_agent.md:11, 31-33, 45-51, 93-107, 147-151`** — **existing prose-level contradiction handling** (Anti-Pattern 3 unresolved contradictions; Step 3 Contradiction Resolution; matrix column "Contradicts"); this sub-issue ADDS structured schema + scoping on top
  - `deep-research/agents/devils_advocate_agent.md` (already does adversarial challenge — relevant but different scope, single-argument not pairwise paper)

#### Problem

`synthesis_agent` already requires identifying and resolving contradictions in prose form (§ Step 3 Contradiction Resolution + Anti-Pattern 3). What is **not yet** present is:

1. **Structured machine-readable output** — contradictions are surfaced as prose narrative, not as `contradiction_pairs[]` schema. Downstream agents (formatter, integrity_verification) cannot programmatically check whether all flagged contradictions have been addressed in the manuscript.
2. **Candidate-pair scoping** — current pass relies on LLM finding contradictions ad-hoc during synthesis. No deterministic scoping of "which pairs to check" — for large corpus (N>30) this becomes inconsistent.
3. **Calibration** — no gold set measuring contradiction detection accuracy.

Note: **#111 `slr_lineage` is a single boolean** (whether systematic-review mode was used) — **not** methodological lineage clusters. Earlier draft mis-cited #111 as candidate-pair scoping base; that was a hallucination caught during review. Candidate scoping must be designed fresh.

#### Proposed change

1. `synthesis_agent` extends existing Step 3 Contradiction Resolution to **also emit structured `contradiction_pairs[]`**:
   ```yaml
[...]

Issue #260, opened 2026-05-28
[Kong sub B1] Experiment Provenance Intake — Material Passport `experiment_provenance[]`


Imbad0202:
> Sub-issue of [META] Kong et al. 2026 (#255): https://github.com/Imbad0202/academic-research-skills/issues/255
> Tier: **B** — Tag: see body below

**Tag**: `[ACCEPT-MEDIUM]` (deep-review finding, P1 severity)

#### Anchor

- **Paper** §3.3 (`kong2026_full.txt:L800-L829`, PDF p.15-16) + §7.4.3 (`L2163-L2176`, p.39): "AI research systems should therefore include explicit evidence ledgers, experiment provenance, versioned manuscript diffs, and revision-tracking mechanisms."
- **ARS files**:
  - `README.md:L109-L123` — current external Experiment Agent flow (user pauses after Stage 1, runs experiment externally, brings results back)
  - `shared/handoff_schemas.md:L481-L507` — current Material Passport schema
  - `.local-plans/2026-04-15-artifact-repro-design.md` — existing `repro_lock` precedent (124 lines, model/prompt/material hashes)
  - `academic-pipeline/agents/integrity_verification_agent.md:L250-L343` — Stage 2.5/4.5 integrity gates

#### Problem

ARS deliberately keeps experiment execution outside the core pipeline (per POSITIONING.md: not auto-research). User runs experiments externally and brings results back. **What is missing is the intake gate**: when external experiment results enter the pipeline, there is no schema-level provenance check tying:

- Repository commit hash for experiment code
- Execution environment (OS, deps, GPU info if relevant)
- Data snapshot ref (S3 path, DVC hash, or equivalent)
- Planned-vs-executed experiment list (which experiments ran, which were skipped)
- Ablation table connecting hyperparams to results
- Negative / null results acknowledged
- Known limitations declared by scholar

Without this gate, downstream `claim_ref_alignment_audit_agent` can verify claim-to-source alignment for cited papers but cannot verify claim-to-experiment alignment for the user's own experiment results.

This is the **provenance intake** layer for external experiments — analogous to how `bibliography_agent` is the provenance intake for external literature.

#### Proposed change

1. Extend Material Passport schema (`shared/handoff_schemas.md`) with optional `experiment_provenance[]` array:
   ```yaml
   experiment_provenance:
     - experiment_id: "exp-001"
       repo_commit: "abc123..."
[...]

PR #521, opened 2026-07-11
Fix/514 add tools allowlist

### LINKED ISSUE #514: Capability allowlists (tools frontmatter) for the three top-level plugin agents — report_compiler_agent has zero guard coverage today
## Motivation

Of ARS's agent surfaces, only the three top-level plugin agents (`agents/synthesis_agent.md`, `agents/research_architect_agent.md`, `agents/report_compiler_agent.md`) are real harness-dispatched subagents; every other agent file is an in-skill prompt template where `tools:` frontmatter would be inert (`scripts/announce-ars-loaded.sh:80,:85`). For those three, frontmatter capability allowlists are a real, harness-enforced boundary — and one of them currently has **zero guard coverage**:

- `report_compiler_agent` is Bucket B: the #294 write-scope guard's Bucket A enforcement does not bind it at all (`scripts/ars_phase_scope_manifest.json:7` documents the Bucket B exemption explicitly).
- For the two Bucket A plugin agents, the guard is an optional Python hook that deliberately fails open when no interpreter is found (`hooks/run_guard.sh:14`). A frontmatter allowlist is defense that survives guard degradation.

## Scope

Narrow — the three top-level plugin agents only, via their canonical `deep-research/agents/` sources + mirror-sync (`scripts/check_agents_mirror_sync.py` pins byte-equality):

1. `research_architect_agent`, `synthesis_agent`: allow required structured write tools, omit Bash (the guard already denies Bucket A Bash when it binds; frontmatter makes that hold when it doesn't).
2. `report_compiler_agent`: omit Bash — the highest-value single change, since it presently bypasses the guard entirely. It keeps structured writes (it produces phase deliverables).
3. Cross-link #330: per-invocation scope grants from the dispatching layer remain the real fix; this is a cheap standing layer underneath, not a substitute.

## Out of scope (deliberate)

- `tools:` frontmatter on in-skill agent .md files — inert, security theater.
- Total write denial for any currently-registered agent — all three produce phase deliverables; none is read-only today. If a genuinely read-only role is ever registered (report returned to a parent that owns persistence), `tools: Read, Grep, Glob` becomes the right shape — that requires real registration + parent-owned persistence, not a frontmatter-only patch.

## Provenance

Mechanism observed in `kengo006/alexandria` (subagent definitions carrying harness-level `tools: Read, Grep, Glob`). Dual-track adversarial in-repo verification (first-party + cross-model xhigh, 2026-07-11): the first-party track initially judged this not worth a standalone issue; the cross-model track's finding that `report_compiler_agent` sits entirely outside the write-scope guard was verified against the manifest and reversed that call, rescoped to the three real plugin agents. Ranked P3 of three evaluated proposals.

madtriceps:
## Eval impact

The eval harness (`.github/workflows/eval-harness.yml`) runs automatically on PRs
that touch scoring / generation logic or the gold sets (see the Delta 3 path
filter in that workflow). Most PRs do not affect eval metrics — leave this
section as "No eval impact." if that applies.

If your change **alters ranking / scoring / generation behavior** and moves a
gold-set metric:

1. Declare each affected metric, one per line, in the exact form:

   ```
   Affected metric: <task>.<class>.<metric>
   ```

   e.g. `Affected metric: citation_extraction.aggregate.accuracy`
   (use class `aggregate` for the headline metric; otherwise the per-class name).

2. If a metric **regresses** (polarity-corrected `signed_lift < -0.05`, or any
   zero-baseline metric changes), the gate blocks unless you add BOTH:

   - the acknowledgement token (on its own line):
     - `[eval-regression-acknowledged]` — for the CI deterministic gate, and/or
     - `[ranking-regression-acknowledged]` — for `scripts/check_ranking_lift.py`
   - a link to an **OPEN** follow-up GitHub issue, e.g.
     `https://github.com/Imbad0202/academic-research-skills/issues/NNN`

> No eval impact.

## Platform port?

If this PR adds a new `<platform>/` directory, read the
[Platform ports policy](https://github.com/Imbad0202/academic-research-skills/blob/main/CONTRIBUTING.md#platform-ports-community-maintained-only)
and **open a design issue first**. Otherwise, leave this section as "Not a platform port."

> Not a platform port.

## Checklist

- [ ] Eval impact section above is accurate

Closes #514

### DIFF
--- a/agents/report_compiler_agent.md
+++ b/agents/report_compiler_agent.md
@@ -2,6 +2,7 @@
 name: report_compiler_agent
 description: "Transforms research findings into polished APA 7.0 academic reports; activated in Phase 4 and Phase 6"
 model: inherit
+tools: Read, Write, Edit, Grep, Glob
 ---
 
 # Report Compiler Agent — APA 7.0 Academic Report Writer
--- a/agents/research_architect_agent.md
+++ b/agents/research_architect_agent.md
@@ -2,6 +2,7 @@
 name: research_architect_agent
 description: "Designs the methodological blueprint; selects research paradigm, method, data strategy, and analytical framework"
 model: inherit
+tools: Read, Write, Edit, Grep, Glob
 ---
 
 # Research Architect Agent — Methodology Blueprint Designer
--- a/agents/synthesis_agent.md
+++ b/agents/synthesis_agent.md
@@ -2,6 +2,7 @@
 name: synthesis_agent
 description: "Integrates findings across sources, resolves evidence conflicts, and maps knowledge gaps"
 model: inherit
+tools: Read, Write, Edit, Grep, Glob
 ---
 
 # Synthesis Agent — Cross-Source Integration & Gap Analysis
--- a/deep-research/agents/report_compiler_agent.md
+++ b/deep-research/agents/report_compiler_agent.md
@@ -2,6 +2,7 @@
 name: report_compiler_agent
 description: "Transforms research findings into polished APA 7.0 academic reports; activated in Phase 4 and Phase 6"
 model: inherit
+tools: Read, Write, Create, Edit, Grep, Glob, WebFetch, WebSearch
 ---
 
 # Report Compiler Agent — APA 7.0 Academic Report Writer
--- a/deep-research/agents/report_compiler_agent.md
+++ b/deep-research/agents/report_compiler_agent.md
@@ -2,7 +2,7 @@
 name: report_compiler_agent
 description: "Transforms research findings into polished APA 7.0 academic reports; activated in Phase 4 and Phase 6"
 model: inherit
-tools: Read, Write, Create, Edit, Grep, Glob, WebFetch, WebSearch
+tools: Read, Write, Edit, Grep, Glob
 ---
 
 # Report Compiler Agent — APA 7.0 Academic Report Writer
--- a/deep-research/agents/research_architect_agent.md
+++ b/deep-research/agents/research_architect_agent.md
@@ -2,6 +2,7 @@
 name: research_architect_agent
 description: "Designs the methodological blueprint; selects research paradigm, method, data strategy, and analytical framework"
 model: inherit
+tools: Read, Write, Create, Edit, Grep, Glob, WebFetch, WebSearch
 ---
 
 # Research Architect Agent — Methodology Blueprint Designer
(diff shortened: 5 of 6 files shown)


Imbad0202:
Hi @madtriceps, thanks for picking up #514. The approach is right: `tools:` frontmatter on the three top-level plugin agents, with Bash omitted, is exactly the intended fix. A few changes are needed before this can merge.

**1. Rebase onto current `main`.** The branch was cut from an older commit, so the diff currently reverts several recently merged changes alongside the fix (the R-CIM rule IDs back to the colliding R-L3-2 namespace, the #260 experiment-backed claims sections, the #262 cross-paper tension inventory, the #518 design-freeze checkpoint, and the phase-boundary enforcement sentence). That is what's failing `spec-consistency`, and it would also break a pytest case (`test_experiment_provenance`). After the rebase, the diff should contain nothing but the frontmatter lines.

**2. Use this exact allowlist:**

```yaml
tools: Read, Write, Edit, Grep, Glob
```

Two adjustments from your current line: `Create` isn't a real tool name (file creation is handled by `Write`), and we're leaving out `WebFetch`/`WebSearch` on least-privilege grounds, since these three agents work from the corpus and prompt context already provided. If a future workflow needs web access for them, we can add it back deliberately.

**3. Apply it to all six files.** The `agents/*.md` copies must stay byte-identical to the `deep-research/agents/*.md` sources, so the same frontmatter line goes into both (`cp deep-research/agents/<name>.md agents/<name>.md` after editing the source).

**4. Add `Closes #514` to the PR body** so the closes-check passes.

Before pushing, please verify locally:

```bash
python3 scripts/check_agents_mirror_sync.py
python3 scripts/check_firm_rules_sync.py
python3 scripts/check_v3_9_2_phase_boundary.py
python -m pytest scripts/test_experiment_provenance.py -q
```

Happy to review again once it's rebased. Thanks again for contributing.

madtriceps:
Hey @Imbad0202 
My bad, i was still working on the PR . Should have marked it as draft. The previous commits are also related to my older commits. Still making a few changes, will open it once ready.
Thanks for review tho.

madtriceps:
@Imbad0202  pls review

Imbad0202:
Merged as 9cdc118. Thanks @madtriceps for the clean rebase and for sticking through both rounds. Re-review came back with zero blocking findings; the one advisory item traced to a collision between two of our own specs, tracked in #523 (nothing on your side), with follow-up hardening in #524.

PR #567, opened 2026-07-19
feat(cross-model): opt-in Codex subscription transport + gpt-5.6-sol promotion bakeoff (#518)

dcs-scd:
## Summary

Adds an **opt-in Codex-subscription transport** for the cross-model citation verifier — ARS's only billable LLM surface — so web-grounded verification can run through a Codex subscription instead of an `OPENAI_API_KEY`, then uses that transport to run the §518 **promotion bakeoff** for `gpt-5.6-sol`.

Commit-by-commit:

1. `d2a9c76` — the transport. `scripts/cross_model_codex_verify.sh` (adapter: isolated `codex exec --json` in a non-git temp dir, `--ignore-user-config --ignore-rules`, fail-closed grounding guard — `searched=true` only on an observed `web_search` item event, `VERIFIED` without ≥1 http(s) URL downgraded to `NOT_SEARCHED`; zero new inline `jq`) + `scripts/cross_model_smoke_test_codex.sh` (mirror of the OpenAI-route smoke test). Auto-wiring is behind a **new opt-in** `ARS_CROSS_MODEL_TRANSPORT=codex` (detection arm in `shared/cross_model_verification.md`, fail-closed to the existing single-model warning; SETUP en/zh-TW quick-setup blocks are byte-identical so the parity lint stays green). Default behavior with the flag unset is byte-equivalent.
2. `569cc22` — the §518 § Promotion Bakeoff, made reproducible: sha256-pinned 30-reference probe set (`fixture_version 2.0.0`), runner/scorer harness, the raw 180-call record, and the audit report (`audits/cross-model-promotion-bakeoff-gpt-5.6-sol-2026-07-16.md`). Scoring is deterministic and offline from the committed record.
3. `4ccf5ea` — `gpt-5.6-sol` **provisional → validated** (#518): all five non-inferiority thresholds pass, no measure close to tolerance. The **recommended default is deliberately NOT flipped** (stays `gpt-5.5`): per §529 a bare non-inferiority pass never flips the default, and the superiority route is not cleanly open (audit Finding 3 — the one discriminating difference is a single reference at Fisher p = 1.0, and raw p95 is nominally behind).
4. `9331790` — `docs/LOCAL_USAGE.md`: end-to-end local guide including a verified known-gaps list.

Two follow-up commits ride on this branch pre-merge (the CHANGELOG coverage key is this PR's number, unmintable earlier): the §518 spec amendment below, and the `[Unreleased]` CHANGELOG entries.

## Review notes — surfaced here deliberately rather than left in the audit

- **Entry gate (audit Finding 7).** §518 names only the API-key smoke test as the promotion entry gate; that script aborts without `OPENAI_API_KEY` and cannot run on this transport. The Codex-route counterpart passed against the candidate on the run date, but its checks 5 (model-id echo) and 6 (effort echo) are **structurally uncoverable** — `codex exec --json` events carry no model/effort field. Check-5 mitigations (verified directly): bogus ids hard-reject with HTTP 400/exit 1 and all 180 calls exited 0; the two arms produce deterministically opposite verdicts on `easy-09`, which one model answering both arms cannot do. This PR amends the §518 spec in-branch to name the Codex-route gate and mark checks 5–6 transport-limited rather than skipped. This was deliberately not patched mid-bakeoff (self-dealing).
- **`easy-09` (audit Findings 2–3).** A reproducible baseline failure: `gpt-5.5` returned `NOT_FOUND` 3/3 with `searched=true` and zero source URLs on a verified-real JACS paper (Jellen 2020, `10.1021/jacs.0c07251`); `gpt-5.6-sol` VERIFIED 3/3 with sources. Replicated 6/6 across two fixture versions. This is exactly the failure class the gate exists to catch, but at n=20 it is one observation (Fisher p = 1.0) and does **not** support flipping the default; the reference stays in the fixture as a permanent canary probe. No larger-n study is planned unless a default flip is on the table.
- **§518 spec defect (audit Finding 6).** Measure 2's ±5 pp tolerance is unexercisable on a 10-item denominator (minimum nonzero delta = 10 pp), so it collapses to "must be ≥ baseline"; measure 3 has exactly one-reference resolution at n=20. Filed in the audit, not fixed here — §532 reserves threshold tuning for a future spec. Happy to open a tracking issue if useful.
- **Deliberate divergence declined.** A review pass suggested the codex smoke test's precondition failures should print `RESULT: FAIL`; rejected to stay faithful to the OpenAI reference smoke test, which uses `ERROR:` + exit 1 on preconditions and reserves `RESULT: FAIL` for check failures.

## Eval impact

The bakeoff artifacts live under a new `evals/bakeoff/` directory — not the gold sets — and no scoring/generation logic is touched.

> No eval impact.

## Platform port?

> Not a platform port.

## Checklist

- [x] Tests added / updated and passing locally — `check_cross_model_verification_sync.py`, `check_setup_cross_model_parity.py`, `check_version_consistency.py` all exit 0; the 5 cross-model-adjacent pytest manifest entries green (174 tests); live smoke test `cross_model_smoke_test_codex.sh` → `RESULT: PASS` (VERIFIED, 3 source URLs, web-search event observed)


Imbad0202:
Thank you for this. I want to be upfront about where I stand: an internal evaluation here had previously set the subscription-CLI route aside, and this PR makes the case I didn't manage to make myself. The fail-closed transport design, the committed 180-call record, and the restraint in the audit (declining the default flip at Fisher p = 1.0 is exactly the right call) won me over on the opt-in form. I want this in.

I ran an independent cross-model review pass over the branch before replying, and it surfaced five things worth discussing. The first two feel merge-gating to me; tell me if you read them differently.

**1. Keeping validation transport-qualified.** All 180 calls ran on the Codex adapter, but the canonical table's `validated` applies to every transport, and API-key users will read it as evidence for a route whose grounding completion, response-shape guards, and latency were never exercised. My suggestion: `validated (codex transport)` in the table, `provisional` retained for the API route, and a line in § Promotion Bakeoff noting an API-route bakeoff completes the promotion. Your audit's transport section already contains all the reasoning the annotation needs; this is a labeling change, not new work.

**2. Non-citation call types.** With `ARS_CROSS_MODEL` set, the Devil's Advocate critique and the two blind disagreement checkpoints also activate, and the citation-specific adapter is the only Codex executor on this branch: a critique carries none of the four verdict tokens and would come back as a successful `NOT_SEARCHED` citation result, which silently swallows those checks. Scoping `ARS_CROSS_MODEL_TRANSPORT=codex` to integrity verification (other call types falling back to the existing single-model warning) seems like the smallest honest fix; a raw pass-through for non-citation prompts would also work if you'd rather keep coverage.

**3. `mapfile` on stock macOS.** macOS ships Bash 3.2, so `cross_model_codex_verify.sh:85` fails there and `set -u` aborts on `verdict_hits`. The repo's other shell surfaces hold a Bash 3.2 line (`announce-ars-loaded.sh`, `hooks/run_guard.sh`); a `while read` loop would keep this one on it.

**4. `CODEX_HOME`.** The smoke test honors a custom `CODEX_HOME` but the detection arm checks only `$HOME/.codex`, so the gate can pass while runtime detection disables cross-model verification. `${CODEX_HOME:-$HOME/.codex}` on both surfaces would align them.

**5. `docs/LOCAL_USAGE.md`.** As committed it describes your fork (your remotes, your branch sequence, the "nothing here is proposed upstream" framing), which stops being true the moment it merges. Dropping it from this PR is fine by me, and the walkthrough can live on in your fork where it's accurate; a repository-neutral rewrite also works if you want it upstream.

Logistics: the branch is a couple of releases behind; I checked a trial merge and only `CHANGELOG.md` conflicts (everything else auto-merges). Happy to handle the rebase on my side after your revision if that's easier. First-contribution CI needs a manual approval on my end; I'll trigger it on your next push.

Independent of all the above, easy-09 is a keeper as a permanent canary, and the audit's own disclosures (the entry-gate limits, the unexercisable tolerance) covered most of what a review round would otherwise have had to surface. Thanks for the work.

Imbad0202:
Thanks for building this. The subscription transport is genuinely useful and I
want it in the suite. On the promotion that rides along with it I have landed
somewhere different, and rather than send you back to redo work I re-ran the
bakeoff myself.

**What my run shows.** Same probe set, unmodified (sha256 `cb7ac8df...`), 30
references x 2 models x 3 repeats, Codex subscription route, 2026-07-30. The
promotion outcome reproduces: all five thresholds pass. Individual numbers differ
from yours, as you would expect across two windows (baseline false-disagreement
and both latency figures moved), so what replicates is the conclusion rather than
the measurements. Everything is on the `bakeoff/518-maintainer-evidence` branch
so you can read the exact commits: the report is at
`audits/cross-model-promotion-bakeoff-gpt-5.6-sol-2026-07-30-maintainer.md` and
the raw record at
`evals/bakeoff/cross_model_promotion/full_run_maintainer_2026-07-30.jsonl`. That
branch vendors your transport and harness verbatim as its base commit, with your
authorship on it, so the diff of my changes reads against what you wrote. It is
not a merge proposal. The
record carries the Codex event stream per call, with one redaction described in
the report and in the transport items below.

**Why I re-ran it, stated fairly.** Not because I doubted your numbers, and not
because you fell short of a rule that existed at the time. The Promotion Bakeoff
section did not ask for timestamps, thread ids, or retained raw output, so the
record you produced met the spec as written. What I found when I tried to check
it is that the spec was underspecified: `run_bakeoff.py` dropped `stdout_raw` on
success, and `score_bakeoff.py` validates nothing about its input, so a
hand-authored JSONL prints `ALL FIVE PASS`. I verified that by writing one. That
is a gap in how I specified the gate, and the fix belongs on my side, which is
where I have put it.

I should be equally straight about the limits of my own run: the new
`audit_run.py` catches incompleteness, not fabrication. A hand-authored record
with plausible thread ids and timestamps still passes all 13 checks, and
`stdout_raw` retains the adapter's JSON rather than the Codex event stream, so
the grounding events are still discarded. My report says so. The honest summary
is contributor result plus independent maintainer replication plus some
repo-side hardening that this exercise prompted, not a failed submission
replaced by a good one.

**One correction, one confirmation, one thing I got wrong.** You are right that
`codex exec --json` carries no model or effort field; I verified it by walking
every key of every event in a live run, and the only model-adjacent key anywhere
is `usage.reasoning_output_tokens`. I had drafted a claim that an invalid model
id does not hard-reject, based on seeing `Model metadata for <id> not found.
Defaulting to fallback metadata`. That was my error: the fallback is client-side
metadata only, and the server still rejects with HTTP 400 and `turn.failed`, exit
1. Your observation was correct. The narrower point I would still make is that a
wrong id failing does not establish which model served a successful call, so the
id check is unmet rather than satisfied by inference.

**One thing that turns out to be checkable.** `auth.json` carries an `auth_mode`
field. The smoke test and the docs check only that the file exists, which cannot
distinguish a ChatGPT subscription from a stored API-key login, so the
subscription claim is currently unattested. Reading `auth_mode` closes that, and
my run records it (`auth_mode: chatgpt`, no API key present).

**On the promotion scope, where I have not finished either.** My run also only
exercised the Codex transport, so it does not by itself support a transport-neutral
`validated`. Three ways out, and I have not picked one: run the API-key route as
well, qualify the status as Codex-transport-validated, or leave the id
`provisional` until both routes are measured. Flagging it because it was a real
gap in the original promotion and my rerun did not close it.

**Items I think need resolving in the transport.**

- `mapfile` is Bash 4. macOS ships Bash 3.2, where it fails with
  `mapfile: command not found`, and under `set -u` the script then aborts before
  emitting any fail-closed JSON. On my machine every call would have returned a
  blank record. A line-oriented count preserves the one-token-once contract.
- `-s read-only` prevents writes, not command execution. This one stopped being
  hypothetical during my rerun: with the event stream retained, 47 of the 180
  calls show the verifier shelling out and reading files from my machine, 192
  `command_execution` items with full file contents in `aggregated_output`. One
  example was `sed -n '1,240p' ~/.codex/skills/.../SKILL.md`. Grounding held (every
  call still completed a web search), but a citation verifier reading the local
  filesystem is not what either of us intended, and with web search available as
  an egress path the shape is worse than the reads themselves. I had to write a
  redaction pass before I could publish the record. I do not have a clean
  acceptance condition to hand you: a minimum-privilege container would satisfy
  me, and so would dropping the contract to advisory, but if neither is reachable
  through the CLI as it stands then that is worth knowing before you spend effort
  on it.
- Grounding is not bound to sources. The check is that some web search completed
  and that the message contains a URL, so a search on something unrelated plus a
  URL echoed from the prompt satisfies it. Without search-result or citation
  binding I read this as a weaker contract than the first-party route, so what I
  would change is describing it as the same one, not necessarily the mechanism.
- The transport selector is not a closed set: any value other than `codex` falls
  through to the API-key path, so a typo with a key present silently switches
  routes. Unset also aborts under `set -u`. The smoke test calls the adapter
  directly, so it would pass even with detection broken.
- Detection and the smoke test disagree on one path: the smoke test honours
  `${CODEX_HOME:-$HOME/.codex}` while detection hardcodes `$HOME/.codex`.
- Availability is declared globally, but the adapter only emits four citation
  verdicts. DA critique and the blind checkpoints cannot round-trip through it.
  Scope availability to citation verification, or implement the other contracts.
- The new parsing logic has no hermetic tests and is not in the pytest manifest.
  Fake-event fixtures covering malformed JSON, no search, wrong event shapes,
  multiple verdict tokens, VERIFIED without a URL, transport failure, and Bash 3.2
  would pin it.
- The PR currently conflicts with `main` and needs a rebase.

**Two scope items.** `docs/LOCAL_USAGE.md` reads as personal working notes (it
says so, and hardcodes your remote, branch and commit), so it is probably better
dropped or rewritten as an operator guide. And the zero-marginal-cost claim is
stronger than I would put it even on a subscription, since usage can draw on
purchased credits.

**Suggested split.** Land the transport on its own once the items above are
settled, with `gpt-5.6-sol` back to `provisional` in this PR. I am asking for the
split for change isolation rather than because the amendment is wrong: a
transport change carrying its own promotion is the structural part, and it is
separable from everything you built here. The 2026-07-19 entry-gate amendment I
agree with substantively and have kept in effect through my own run; when it
lands it should keep your name on it, and it should also absorb the model-id
finding above.

Your bakeoff design is sound and I kept all of it: the probe set unmodified, the
three-repeat majority rule, and the conservative treatment of indeterminate
results. The harness changes I made were about keeping receipts, not about what
is measured. Thank you for doing the design work that made a maintainer run
possible at all.

Imbad0202:
Superseded by #705 for #630.

#705 preserves the subscription-transport requirement and credits @dcs-scd in the design, changelog, PR description, and commit co-author trailer. It narrows the feature to citation-integrity calls and replaces `codex exec --json` with app-server v2 structured `webSearch.results[]` binding, so a returned URL must trace to the exact result seen by the model. The promotion/bakeoff and personal-operator-guide portions remain intentionally out of scope.

Thank you for the original implementation and product requirement.

Issue #593, opened 2026-07-26
Proposal: a clinical-medicine module track — EQUATOR reporting-standard gates, medical disclosure venues (incl. Chinese Medical Association policy), and a Chinese-literature citation resolver


ktao732084-arch:
Hi — first, thank you for building ARS the way you did. We spent real time reading the code (not just the README): the fail-closed discipline and the three-way `matched/unmatched/unresolvable` verdicts in `scripts/verification_gate/`, the typed-hit validation and staleness advisories on the verification cache (#557), the `gold/` vs `heldout/` eval split, the grounded cross-model checks, and the `compliance_agent` framework running PRISMA-trAIce + RAISE at the 2.5/4.5 integrity gates. It is the most engineering-mature project of its kind we found, and it is motivated by the same evidence we care about (Zhao et al. arXiv:2605.07723 on hallucinated citations).

We're a small **non-commercial** effort building AI-assisted-research support for **clinical medicine** (our end users are hospital clinicians, mostly writing in Chinese). Rather than rebuild a pipeline that already exists and is better than what we'd make, we'd like to contribute a clinical/medical discipline module upstream — CONTRIBUTING.md's "discipline-specific modules are encouraged / domain-agnostic" line and the #259 domain-evidence-profiles direction are what prompted this. CC BY-NC 4.0 fits us fine (we're non-commercial too), and we intend to meet the engineering bar — tests, manifest registration, CHANGELOG discipline, README sync.

### Directly PR-able (maintainer-review tier) — each arrives as its own small PR with its own issue

[...]

Issue #495, opened 2026-07-06
design(literature): API-first retrieval with browser-controlled fallback


pikaqiu2333:
## Motivation

ARS already has strong structured citation/literature verification surfaces:

- `deep-research/references/semantic_scholar_api_protocol.md`
- `deep-research/references/openalex_api_protocol.md`
- `deep-research/references/arxiv_api_protocol.md`
- WebSearch/WebFetch-based human-readable verification paths

I would like to propose a small protocol refresh that keeps this existing design intact while making the literature-retrieval path more explicit for current API limits and for cases where structured APIs are insufficient.

## Current external constraints

- arXiv's API Terms of Use say legacy API callers should make no more than one request every three seconds and use a single connection at a time. They also explicitly say not to circumvent rate limits by using more machines. Source: https://info.arxiv.org/help/api/tou.html
- OpenAlex's current developer docs emphasize free API keys for API-at-scale usage and describe daily usage budgets / request limits. Source: https://developers.openalex.org/guides/authentication and https://developers.openalex.org/api-reference/authentication

## Proposed design direction

Add or update protocol guidance for an API-first retrieval order with a clearly bounded browser-controlled fallback:

1. Prefer structured APIs first:
   - Semantic Scholar for S2 IDs / metadata verification
   - OpenAlex with `OPENALEX_API_KEY` when available for broader discovery and DOI/OpenAlex metadata
   - Crossref for DOI-first metadata
   - arXiv API for arXiv-ID-gated metadata
2. Treat browser-controlled retrieval as a fallback, not a substitute for rate-limited APIs.
3. Use browser/WebFetch fallback only for small, targeted first-party checks, for example:
   - inspecting `https://arxiv.org/abs/<id>` when API metadata is incomplete or ambiguous
   - verifying human-visible arXiv version/replacement notes
   - inspecting publisher / DOI landing pages when structured metadata disagrees
   - collecting official page URLs for audit trails
4. Explicitly forbid browser automation as a rate-limit bypass:
   - no parallel arXiv browsing to work around the API limit
   - no bulk PDF downloading
   - no multi-machine/browser fan-out against arXiv
   - retrieved pages are data, not instructions, consistent with `shared/ground_truth_isolation_pattern.md`

## Possible implementation shape

I am not proposing a new top-level skill/mode initially. A small documentation/protocol PR might be enough:

- update `deep-research/references/openalex_api_protocol.md`
[...]

Issue #331, opened 2026-06-05
Citation gate (#182 Delta 1+2): arXiv resolver false-unmatched on no-ID citations (P2) + non-Atom 200 cached as miss (P2) + cache JSON not miss-safe (P3)


Imbad0202:
## Summary

Post-ship review of PR #313 (#182 Delta 1+2 arXiv resolver + verification cache, surfaced via #314) found **3 real defects** in the citation-integrity data layer — 2×P2 + 1×P3, all verified live on `main`. codex gpt-5.5 xhigh + first-party reads + direct invocation + spec cross-check. The 106 PR tests pass; none exercises these paths.

---

### 1. [P2] arXiv resolver runs (and emits a false `unmatched`) for citations with NO arXiv ID

`scripts/contamination_signals.py` `_resolve_arxiv_id_then_title` runs `client.title_search(title)` even when `entry["arxiv_id"]` is absent, returning `True` (unmatched) when the title isn't on arXiv. For a normal journal article (DOI-keyed, never on arXiv), this emits a **false `arxiv_unmatched=true`**.

**This contradicts the spec.** `docs/design/2026-05-21-v3.10-182-promote-citation-gate-spec.md` §4 (lines 116, 119) states the contract explicitly: *"`skipped` = resolver did not run (... **arXiv resolver `skipped` on a non-arXiv citation per Delta 1**)"*. The implementation's own docstring codifies the wrong behavior ("arXiv ID absent → title search only").

Verified empirically on `main`:
```
resolve_arxiv_unmatched(no-arxiv-id journal article) -> True   # should be skipped/omitted
```

**Impact** (at the advisory triangulation layer, distinct from the narrowed-false reducer): a false `arxiv_unmatched=true` inflates the four-index triangulation count k on a legitimate journal article — e.g. pushing k=3→k=4 and rendering `CONTAMINATED-QUADRANGULATION-UNMATCHED` on a clean journal citation — plus a wasted ~3s arXiv request per such citation. The narrowed-false `lookup_verified` gate is protected (a title-only unmatched reduces to `unresolvable` per C-V6(a)), but the contamination/k advisory surface is not.

**Fix:** short-circuit `_resolve_arxiv_id_then_title` to return the `skipped` outcome when `arxiv_id` is absent — do not run the title search. Update the docstring. (#315's `queried_by` refactor touched this function's return tuple but not the no-ID skip.)

---

[...]

PR #600, opened 2026-07-26
feat(scripts): Chinese-literature resolver client + API protocol doc

### LINKED ISSUE #595: Chinese-literature citation resolver: standalone client + API protocol doc (gate untouched)
> Part of the clinical-medicine module track proposed in the umbrella issue #593. Standalone client only — no gate, schema, summary, or triangulation-matrix changes (that integration is deliberately issue-first, per the umbrella).

## The underlying fact: Crossref is one DOI registration agency, not the DOI system

Real, resolvable Chinese-literature DOIs return 404 from the Crossref API because they are registered with a different registration agency (ISTIC or CNKI). Reproducible in seconds:

```
$ curl -s https://doi.org/ra/10.3760
[{"DOI": "10.3760", "RA": "ISTIC"}]        # Chinese Medical Association journal prefix

$ curl -s -o /dev/null -w "%{http_code}" https://api.crossref.org/works/10.3760/cma.j.cn112137-20240101-00001
404                                         # Crossref API: not found

$ curl -sL -H "Accept: application/vnd.citationstyles.csl+json" https://doi.org/10.3760/...
200 + CSL-JSON with the original Chinese title, journal, volume/issue/pages
```

(Illustrative DOI shown; the PR's protocol doc will carry hand-verified live examples.) `https://doi.org/ra/<prefix>` triages the registration agency at zero cost and supports comma-batching; content negotiation against `doi.org` follows the RA's metadata service, which for ISTIC returns usable CSL-JSON.

For a manuscript citing Chinese core-journal literature, "not found in Crossref (or OpenAlex/S2, whose Chinese-journal coverage is thin — OpenAlex indexes ~37% of Chinese core-list journals and ~24% of their articles, arXiv:2512.16339)" is therefore weak evidence about existence in either direction. This matters for the precision-over-recall promise: a fabricated Chinese citation can often be *cleanly* refuted (Handle API has no wildcard catch-all — a made-up DOI returns `responseCode: 100`), while a real one currently looks identical to a miss.

## Proposal: `scripts/` Chinese-literature resolver client, mirroring the arXiv client spec

A standalone client + API protocol reference doc + transport-fixture tests, following the #495/#496 community precedent:

- **Interface discipline mirrors `arxiv_client.py`:** typed `Unavailable` exception (caller must degrade explicitly, never fabricate a verdict), per-instance `time.monotonic` throttling, no-retry-on-5xx, exact-title-or-bust cross-checking (with NFKC + CJK punctuation/whitespace normalization — we measured fuzzy title matching to be *more* dangerous for Chinese titles than English ones).
- **Resolution waterfall, all lawful open interfaces, no CNKI/Wanfang scraping:** applicability check (non-Chinese → skipped) → DOI present? → RA triage via `doi.org/ra/` → ISTIC: doi.org content negotiation + title cross-check; CNKI-registered: Handle existence only (their RA serves no CSL-JSON; we deliberately refuse to parse resolution HTML) → no DOI: journal-title → ISSN → NLM TA bridge, PubMed coverage confirmation ("journal has ≥1 PubMed record", not "journal is in the NLM catalog" — those differ in practice), then E-utilities coordinate query (`[ta]+[vi]+[pg]`).
- **Asymmetric confidence by design:** negatives are strong (refuted DOI / title mismatch on an ID-keyed lookup), positives are conservative — "DOI exists but title unverifiable" is never a match. Everything unresolved lands in a **human-confirmation list** (prioritized P0–P3), never a "fabricated" verdict — consistent with C-V6(a)'s ID-keyed narrowing and the `unresolvable` philosophy.

## Scope

- **In:** `scripts/<name>_client.py`, `deep-research/references/<name>_api_protocol.md`, unit + transport-fixture tests (fully synthetic bodies, no live network in CI), pytest-manifest registration, CHANGELOG entry.
- **Out (explicitly):** `verification_gate`, `citation_verification_summary` (pinned to exactly four `resolver_outcomes`), the k=0..4 triangulation matrix, suffix lints, degradation registry — untouched. Gate integration is the issue-first item in #593, only if/when you want it.

ktao732084-arch:
## Summary

Standalone client for verifying Chinese-literature citations, closing the gap demonstrated in #595: `api.crossref.org` is one DOI registration agency, not the DOI system — real Chinese DOIs (ISTIC / CNKI-registered) 404 on the Crossref API yet resolve normally through `doi.org`, so the existing four resolvers reduce nearly every Chinese reference to the same `unresolvable` state a fabricated one produces.

- `scripts/chinese_literature_client.py` — waterfall over four open, key-free upstreams (no CNKI/Wanfang/VIP scraping, pinned by an AST-level test): `doi.org/ra/` RA triage → ISTIC content negotiation + Chinese-aware exact-title matching / CNKI Handle existence probe only → ISSN → NLM-TA bridge → PubMed coverage confirmation → `[ta]+[vi]+[pg]` coordinate query. Interface discipline mirrors `arxiv_client.py` (typed Unavailable, monotonic pacing, 429 backoff at endpoint floor, no 5xx retry, #331 guard), plus the #495-style key protections: a three-host allowlist guard, and error-path URL redaction so an NCBI `api_key` can never land in raised-exception text (pinned by tests across the 5xx / network-error / bad-body paths).
- **Precision asymmetry by design:** only a refuted DOI (Handle responseCode 100 — no wildcard catch-all exists on these prefixes) or an ID-keyed title mismatch produces the C-V6(a) shape that licenses `false`; every other non-match stays title-keyed → `unresolvable`, with signal strength expressed as a P0–P3 human-confirmation checklist instead of a verdict. `resolve()` has no `unreachable` return — degradation raises, the caller decides.
- **A measurement drove one departure from shared helpers:** on CJK titles the shared fuzzy ratio scores an *unrelated* paper 0.510 vs 0.577 for a fullwidth spelling of the *identical* title — ANDing it in demoted a correct match to a P0 "refuted" row during live smoke testing (the exact false-accusation mode this design exists to prevent). The matcher is therefore exact-normalized equality (NFKC + Unicode punctuation-category stripping + whitespace removal); the measurement table is in the protocol doc and pinned by regression tests.
- `deep-research/references/chinese_literature_api_protocol.md` — endpoints, rate-limit etiquette, degradation table, three-state semantics ("why a miss is never *fabricated*"), legal boundary, hand-verified live examples (re-verified 2026-07-27), and an explicit list of the three schema-side deltas a future gate wiring would need (`resolver_outcomes` is pinned to four keys; `queried_by`/`skipped` semantics) — the status/queried_by vocabulary itself matches `citation_verification_summary.py` verbatim.
- 48 tests over fully synthetic transport bodies (reserved `10.5555` prefix / out-of-range PMIDs / `.invalid` hosts; README documents the synthetic-only discipline); zero live network in CI; pytest manifest registered.

Deliberately standalone: `verification_gate`, `citation_verification_summary`, the triangulation matrix, schemas, and the degradation registry are untouched — gate integration is the issue-first item in #593 and only happens if you want it.

Verification: full local suite 3999 → 4047 passed (+48), 3 skipped / 1 xfailed unchanged. Full local run of the spec-consistency lint battery passes. Local live smoke (not in CI): 11/11 scenarios behave as designed, including a real NCBI transient 500 correctly raising instead of minting a verdict.

Closes #595. Part of the #593 clinical-medicine track.

## Eval impact

> No eval impact.

(Checked against the Delta 3 path filter explicitly: `chinese_literature_client.py` is a new file not listed there, and no listed path is touched.)

## Platform port?

> Not a platform port.

## Checklist

- [x] Tests added / updated and passing locally (+48, full suite green)


Imbad0202:
Thanks for this, and the underlying observation is correct and worth stating
plainly: Crossref is one registration agency, not the DOI system. I reproduced it
before reviewing. `https://doi.org/ra/10.3760` returns ISTIC, and a real
PubMed-indexed `10.3760/cma.j.cn112137-...` DOI returns 404 from the Crossref API
while the Handle API confirms it exists (`responseCode` 1, against 100 for one I
made up). That asymmetry is real and worth a resolver.

The scope promise holds: no `verification_gate/`, no summary schema, no
triangulation matrix, no degradation registry. The pytest manifest registration
and CHANGELOG entry are in-scope sync. 48 tests pass at the PR head.

What I would like to resolve before it lands is one family of issues: a handful of
paths that can produce a confident wrong answer, which is the failure mode the
design itself sets out to avoid. I suspect most of these are cases where the
implementation drifted from the intent stated in #595.

**Redirects leave the allowlist.** The three-host HTTPS allowlist is checked on
the initial URL only, and `urlopen` follows redirects. The documented ISTIC DOI
currently redirects to `http://122.115.55.36:8000/...`: plaintext, a bare IP,
outside the allowlist. Metadata from that hop decides MATCH versus
DOI_TITLE_MISMATCH. The unbounded `resp.read()` on the same path compounds it.
Could the client validate the host on every hop, refuse an HTTPS-to-HTTP
downgrade, and bound the body? And if ISTIC's metadata turns out to be reachable
only over plaintext, my instinct is that branch should stop short of an automatic
verdict rather than trust it. In fairness this is shared debt rather than
something you introduced: the existing clients validate the allowlist before
redirects too and read responses unbounded. What I want to avoid is adding a new
verdict path over plaintext; I will open a follow-up for the ones already in
tree.

**The normalizer collides on scientifically distinct titles.** Stripping every
Unicode `P*` and `S*` character makes different papers compare equal. Verified
against the implementation in this PR:

| title A | title B | normalized |
|---|---|---|
| `ER+ 乳腺癌患者的预后分析` | `ER- 乳腺癌患者的预后分析` | identical |
| `CD4+ T细胞计数与预后` | `CD4− T细胞计数与预后` | identical |
| `生存率提高 4.5% 的临床意义` | `生存率提高 45% 的临床意义` | identical |

Operators, signs and decimal points carry meaning in scientific titles generally,
not only clinical ones. (These are constructed pairs, not two papers I found in
the wild, but they are the shape a real collision would take.) Stripping only characters shown to be semantically inert,
plus collision tests for these cases, should close it.

**Other paths that can produce a wrong verdict:**

- The `generic_title` veto is applied to ID-keyed DOI lookups. A real DOI whose
  title legitimately is "Editorial" gets vetoed, then Handle-exists turns that
  into DOI_TITLE_MISMATCH and a `false`. The sibling clients only use that veto
  on title searches without an identifier.
- A 404 from the CSL endpoint and a genuine title mismatch both collapse to
  `None`, so a DOI that exists but whose title was never retrieved is reported as
  a title mismatch. Those need to stay distinguishable: not-found, mismatched,
  and unverifiable are three different states.
- A unique PubMed coordinate hit is accepted after checking ISSN and year only,
  with no binding back to the cited title. Substituting an invented Chinese title
  into the existing test still yields PUBMED_COORDINATE_VERIFIED. A coordinate hit
  should either confirm the title through the record's DOI or stay a human check.
- The author fallback uses `[au]`, but PubMed's first-author tag is `[1au]`, so a
  different paper by the same co-author can be selected. Volume, pages, author and
  year are interpolated into Entrez grammar, where URL encoding does not neutralise
  field tags or boolean operators.
- A DOI whose registration agency is Crossref or unknown falls through to the
  coordinate branch that the protocol doc says is for references without a DOI. A
  wrong DOI plus real coordinates then reaches `matched`.

**One documentation point.** The motivation says the existing four resolvers
reduce nearly every Chinese reference to `unresolvable`. I started to push back on
that, because OpenAlex does return a full record for the protocol doc's own
example DOI, then checked what our resolver would do with it: OpenAlex serves an
English shadow title, which scores 0.202 against the Chinese original, far under
the 0.70 floor, and fails exact-normalized matching outright. So the record
existing upstream does not make the reference resolvable here, and your claim
stands as written. Worth stating that mechanism explicitly in the protocol doc,
since "OpenAlex has it" is the first objection anyone will raise. Separately, the DOI Foundation documents `/doiRA/`
rather than `/ra/` as the registration-agency endpoint; `/ra/` works today, but
"all documented interfaces" should use the documented one.

**Smaller items.** ISTIC returning HTTP 200 with a non-JSON body becomes a
routine human-check row rather than a typed `Unavailable`, which is not the
fail-closed behaviour the PR describes; structurally malformed but parseable JSON
escapes the typed degradation contract with a bare `TypeError`; the PubMed branch
reads `container_title`, `volume`, `pages` and `first_author_pinyin`, which the
current corpus-entry schema does not carry, so that input-contract change belongs
in the integration-delta list; NCBI requests should carry the `tool` and `email`
parameters; the fixture README's claim that `10.5555` is structurally unmintable
is not right (Crossref registers resolvable DOIs under it); and the CSTR exclusion
is fine but the stated reason is not (it has an authenticated detail endpoint, so
"not key-free" is the accurate reason).

**Attribution.** Per CONTRIBUTING.md's Credit section, the CHANGELOG entry should
credit you by handle and the README Contributors section should be updated. That
is required sync, not scope creep. Same applies to the other two PRs.

Also needs a rebase; it conflicts with `main` on CHANGELOG and the pytest
manifest.

The shape of this contribution is one I would not want changed: standalone
client, protocol doc, synthetic transport fixtures, no live network in CI, gate
untouched. With the paths above tightened, this is the kind of resolver the suite
is missing, and I would like to get it in.

ktao732084-arch:
Thanks for the detailed review. I have pushed the revision at exact head `532a0d6f`, rebased onto current `main` (`462b32b`).

The updated resolver now:

- validates HTTPS, the exact allowlist, DNS/IP resolution, and response-size bounds on every redirect hop, and stops before any verdict on the current plaintext bare-IP ISTIC redirect;
- preserves semantically meaningful scientific-title characters such as `+`, `−`, `%`, and decimal points, with collision regressions;
- keeps DOI results as `MATCH / MISMATCH / NOT_FOUND / UNVERIFIABLE`, separates Handle existence from title verification, and prevents a wrong DOI from falling through to coordinate matching;
- treats PubMed coordinate hits as candidates only, requiring exact PMID/DOI echoes and machine-readable Chinese-title rebinding before `matched`;
- uses the first-author field safely, closes Entrez grammar-injection and malformed-payload paths, and keeps all degradation typed and fail closed;
- synchronizes the corpus-entry integration delta, NCBI metadata, documented DOI RA endpoint, fixture provenance, CHANGELOG, and contributor credit.

The contribution remains standalone: no production verification gate, triangulation matrix, schema gate, or degradation-registry integration was added.

Verification on `532a0d6f`:

```
167 focused tests passed under Python 3.9 and 3.12 with warnings treated as errors
4942 passed, 3 skipped, 1 xfailed, 140 subtests passed
```

Manifest, spec/version-consistency, compile, and diff checks pass. Independent network-boundary and verdict-state-machine audits found no remaining P0–P2 issues.

Ready for re-review at exact head `532a0d6f`.

Imbad0202 (approved):
Exact-head maintainer review complete: two independent review tracks clean; 174 focused tests and 5,174-test full suite passed locally; all GitHub checks green.

Issue #512, opened 2026-07-11
PDF read-integrity preflight for locally-extracted page/quote anchors (upstream truncation/mispagination guard for Three-Layer emission)


Imbad0202:
## Motivation

The v3.7.3 Three-Layer Citation Emission guards locator **presence**, and the #182 gate guards citation **existence** — but nothing guards the **local extraction channel** those locators come from. Harness/tool PDF readers can silently truncate documents with malformed cross-reference tables and misreport page counts; a real, correctly-cited source can then acquire an apparently valid `page` anchor derived from a truncated or mispaginated read, and pass every current gate:

- The three emitters anchor solely from corpus context already in their prompt and are forbidden from re-reading entry frontmatter (`deep-research/agents/synthesis_agent.md:293` R-L3-1-C; same rule in `draft_writer_agent` / `report_compiler_agent`) — poisoned context upstream means poisoned anchors downstream, emitted in good faith.
- `scripts/check_v3_7_3_three_layer_citation.py` checks anchor shape and quote length, not faithfulness to the source.
- The #182 gate reduces the anchor to a kind-only boolean (`scripts/verification_gate/__init__.py:123` `_anchor_present`); it never checks the value or the PDF.
- The finalizer's 4-cell matrix + precedence-zero locator check (`academic-pipeline/agents/pipeline_orchestrator_agent.md:691,:717`) can resolve `ok` without any read-integrity evidence.
- One script already reads user PDFs (`scripts/bootstrap_timeline_yaml.py:48`, `pdftotext` on the cover page), so a preflight script is not a new category of tooling for this repo.

## Scope

Two layers, enforcement upstream:

1. **PDF-integrity preflight script** invoked at the orchestration/retrieval layer (not by writer agents — Bucket A agents cannot run Bash). Compare the root page-tree `/Count` against an independently enumerated page count; surface xref-repair/parser warnings. Not "grep the first `/Count`" — handle nested page trees, incremental updates, object streams. Emit `PASS / FAIL / UNAVAILABLE` + file hash + declared vs enumerated counts in a sidecar/retrieval result.
2. **Prompt rules** in the three emitters + the L3 audit agent: a `page` anchor from a locally-read PDF requires preflight `PASS`; on `FAIL`/`UNAVAILABLE`, emit `anchor:none` (letting the existing gate surface it) or a non-page locator whose text is independently visible, with an explicit PDF-integrity warning.

## Out of scope (deliberate)

[...]

PR #601, opened 2026-07-26
docs(deep-research): add CARE / STARD 2015 / TRIPOD+AI condensed guidance and a study-design routing sequence

ktao732084-arch:
## Summary

`equator_reporting_guidelines.md` carried condensed guidance for PRISMA / CONSORT / STROBE / COREQ / SQUIRE, while the three study types a clinician-author most often writes first — case reports, diagnostic accuracy studies, prediction models — were name-only rows in the mapping table. This PR extends the file in its existing shape:

- **§7 CARE** (13 topics / 30 items), **§8 STARD 2015** (all 30 official item ids), **§9 TRIPOD+AI** (all 27 official item ids) — item-level guidance keyed to the official item ids, source-statement citation with DOI, official checklist URL, and Clinical Research Application Notes parallel to the existing Higher-Education notes.
- **§10 Study Design → Reporting Guideline Routing** — the ordered question sequence the mapping table previously assumed had already happened (original research → investigator-assigned intervention → randomised → single case → index test vs reference standard → prediction model → sampling direction), with companion-tool overlays (RECORD / COREQ / CHEERS / ARRIVE) and a load-bearing **ambiguity rule**: an ambiguous description is a question for the scholar; a study design is never inferred from keywords. A short table of common manuscript phrasings that do *not* settle the routing (e.g. "randomly divided into two groups") records the follow-up question each should trigger instead.
- Mapping table rows bolded/updated accordingly; the old §7 (Higher Education recommendations) renumbers to §11 — we checked every in-file cross-reference and swept the repo for external section-number references to this file (there are none; all six consumers reference it by filename only).

**Copyright:** no checklist text is transcribed — all item wording is ARS paraphrase, stated in a blockquote in each new section. The licensing situation is non-uniform per statement (CARE's official site is non-commercial-licensed; the STROBE site PDF carries no licence statement; STARD 2015 and TRIPOD+AI are CC BY), so paraphrase-with-attribution is the only uniformly safe form, and it matches the file's pre-existing style. We verified the paraphrase claim mechanically: every table fragment scores below 0.85 against the official item wording under `scripts/_text_similarity.py`.

[...]

PR #623, opened 2026-08-02
Advisory PDF content classification (scanned/OCR detection) for the #512 read-integrity preflight

gianpierreperu:
## Summary

Adds an optional, advisory-only content classification signal to the #512 PDF read-integrity preflight (`scripts/pdf_read_preflight.py`), using the [pdf-inspector](https://github.com/firecrawl/pdf-inspector) package (Rust-backed, MIT, no OCR required) to detect scanned/image-only pages that #512's structural checks cannot see.

**Motivation:** #512 verifies page-tree *structure* (three independent page-count signals must agree) before trusting a locally-extracted `page` anchor. But a scanned or image-only PDF can carry perfectly consistent, agreeing page counts while yielding empty or meaningless extracted text — a structural `PASS` today says nothing about whether the underlying text a quote/page anchor was minted from was ever real. This was explicitly out of #512's scope (its design doc's "Out of scope" section defers quote-accuracy verification to the L3 claim-audit channel — this sits one layer earlier, at "was there text to read at all").

**What's included:**

- `_classify_content()` in `scripts/pdf_read_preflight.py`: runs `pdf_inspector.classify_pdf_bytes()` independently of and before the pypdf-based structural checks, adding a `content_classification` field (`pdf_type` / `confidence` / `pages_needing_ocr`) to the sidecar.
- Strictly advisory — verified by inspection and pinned by test that `result["verdict"]` (PASS/FAIL/UNAVAILABLE) is governed only by the pre-existing three-count-agreement / parser-warning / trailing-data logic; this addition only appends to `warnings`, which the verdict logic never reads.
- Graceful degradation tuned for an optional dependency: plain absence of `pdf_inspector` is silent (no warning — the common default state), while an installed-but-erroring classifier does warn.
- New `_flat_text_pdf()` test fixture (real `Tj` text content) so the existing "clean PDF, zero warnings" test stays meaningful when `pdf_inspector` is genuinely installed, without touching the ~30 pre-existing structural-only tests that intentionally use content-free fixtures.
- Design doc: `docs/design/2026-08-01-pdf-content-classification-spec.md`, with an explicit Out-of-scope section (no Layer 2 / agent-prompt wiring, no OCR routing, no quote-accuracy verification — left for a separate, maintainer-discussed change per the agent-definition-change bar in CONTRIBUTING.md).
- CHANGELOG entry under `[Unreleased]`.

**Process note (transparency):** this patch was developed and self-reviewed outside the normal issue-first flow for maintainer-reviewed changes — no issue was opened before implementation. An independent adversarial code review (a second pass with no prior context, not this repo's cross-model process) caught one real bug before this PR was opened: an earlier revision's `try/except` around `pdf_inspector.classify_pdf_bytes()` only wrapped the call itself, leaving attribute access on the returned object unguarded — a future/beta library version returning an object with a missing or renamed attribute would have raised an uncaught exception straight through `run_preflight()`. Fixed, with a regression test added (a stub object with no attributes at all). Flagging this proactively.

**Tests:** `pytest scripts/test_pdf_read_preflight.py` — 37/37 pass, both with `pdf_inspector` absent (default) and genuinely installed (`pdf-inspector==0.2.6`). `python scripts/check_version_consistency.py` passes.

I'd expect this falls under "Requires maintainer review" (new capability, not a bug/drift fix) per CONTRIBUTING.md — happy to open a design issue first instead if you'd rather discuss scope before reviewing code, or to split this differently.

## Eval impact

The eval harness (`.github/workflows/eval-harness.yml`) runs automatically on PRs
that touch scoring / generation logic or the gold sets (see the Delta 3 path
filter in that workflow). Most PRs do not affect eval metrics — leave this
section as "No eval impact." if that applies.

If your change **alters ranking / scoring / generation behavior** and moves a
gold-set metric:

1. Declare each affected metric, one per line, in the exact form:
Affected metric: <task>.<class>.<metric>

e.g. `Affected metric: citation_extraction.aggregate.accuracy`
(use class `aggregate` for the headline metric; otherwise the per-class name).
2. If a metric **regresses** (polarity-corrected `signed_lift < -0.05`, or any
zero-baseline metric changes), the gate blocks unless you add BOTH:
- the acknowledgement token (on its own line):
  - `[eval-regression-acknowledged]` — for the CI deterministic gate, and/or
  - `[ranking-regression-acknowledged]` — for `scripts/check_ranking_lift.py`
- a link to an **OPEN** follow-up GitHub issue, e.g.
  `https://github.com/Imbad0202/academic-research-skills/issues/NNN`
> No eval impact.
## Platform port?

If this PR adds a new `<platform>/` directory, read the
[Platform ports policy](https://github.com/Imbad0202/academic-research-skills/blob/main/CONTRIBUTING.md#platform-ports-community-maintained-only)
and **open a design issue first**. Otherwise, leave this section as "Not a platform port."
> Not a platform port.
## Checklist


Imbad0202:
Thanks for the contribution. The underlying gap is real: #512 validates PDF page-tree structure, not whether usable text actually exists. I also checked the current open issues and PRs; this is complementary to #512 and does not duplicate the ongoing work.

That said, I do not think the patch is ready to merge in its current shape. Please address these points before another review:

1. **Open and link a dedicated design issue.** This is a new, citation-integrity-adjacent capability, and the PR acknowledges that it skipped the repository's issue-first flow. A linked issue is also needed for the repository's closes-issue coverage workflow.

2. **Define the consumer and observable behavior.** Today the #512 consumers inspect the sidecar's `verdict`; none of them consumes `content_classification` or its warning. This PR deliberately leaves the verdict unchanged and does not wire the signal into the orchestrator, emitters, claim-audit finalizer, OCR routing, or anchor policy. Consequently, a scanned PDF still receives the same structural `PASS`, and the new value is currently unconsumed metadata. Please either:
   - scope this explicitly as a diagnostic-only facility and identify its concrete consumer/use path, or
   - propose the downstream advisory/routing behavior in the issue-first design discussion.

   Until a consumer exists, please avoid framing this as closing the reliability gap end to end.

3. **Make the optional-dependency boundary consistent.** The import path treats `pdf_inspector` as optional and silently degrades when absent, but `requirements-dev.txt` installs it unconditionally. Please clarify the supported installation model and test both dependency-present and dependency-absent paths in the intended CI matrix.

4. **Preserve the preflight's failure-containment guarantee.** `pdf-inspector` is Rust/PyO3-backed. `except Exception` does not cover PyO3's `PanicException` (a `BaseException` subclass), and no in-process handler can recover from a native abort or segmentation fault. Since #512 relies on the preflight returning a sidecar rather than taking down PDF intake, please isolate classification behind a subprocess boundary (with timeout, exit-status handling, and validated output), or provide an equivalent demonstrated containment mechanism. Tests should cover timeout, non-zero exit, and malformed output as advisory degradation.

The direction is potentially useful, and there is no functional conflict with #567, #600, or #601 beyond the shared `CHANGELOG.md`. The main issue is that this currently adds execution/dependency risk to every local-PDF preflight without a defined downstream consumer.

Imbad0202:
Thank you again for raising this. PR #724 was merged in v3.20 and now provides the opt-in, process-isolated PDF text/OCR advisory proposed here, while preserving the structural verdict, keeping `pdf-inspector` optional, and covering dependency absence, timeout, nonzero exit, malformed output, and schema validation. The shipped design names itself as the safe replacement for the concept in this PR, so I am closing it as superseded. Automatic OCR routing and anchor-policy changes remain separate issue-first work.

PR #636, opened 2026-08-03
feat(platform): add community-maintained Pi wrapper

### LINKED ISSUE #631: Design: Pi Agent in-tree wrapper
I propose adding a small, community-maintained Pi Agent wrapper under `pi/`. I would prefer an in-tree wrapper because it keeps the original ARS skills, agents, references, scripts, and commands as the only source of truth.

No ARS prompt content is vendored or manually synchronized. Changes to the original skills become available to Pi automatically after updating the repository. It also avoids maintaining a second repository and a duplicated release process.

Because this is an in-tree wrapper, ARS workflow content is not vendored. Pi-specific issues can be assigned or redirected to me / @HannesOberreiter.

## Current Prototype:

Prototype currently tested with:

- `@earendil-works/pi-coding-agent` 0.83.0
- provider: `openai-codex`
- model: `gpt-5.6-sol`

### Files

```text
package.json            # minimal Pi Git-install entry point
pi/
  package.json          # wrapper-scoped manifest and local installation
  wrapper.js            # dependency-free Pi adapter
  README.md             # installation, behavior, and limitations
examples/
  pi/                   # full academic-pipeline evidence before PR
```

Core ARS files would remain unmodified, including:

The root `package.json` only enables direct installation from GitHub:

```bash
pi install git:github.com/Imbad0202/academic-research-skills
```

The actual adapter remains under `pi/`. The root manifest has no runtime dependencies. The wrapper does not copy or rewrite their `SKILL.md`, agent, reference, schema, or script content.

### Original commands

All original `commands/ars-*.md` files remain authoritative.

When an `/ars-*` command is invoked, the adapter:

1. reads the original Claude command file at invocation time
2. removes Claude-specific frontmatter
3. preserves user arguments
4. routes mode commands through Pi's native `/skill:*` mechanism
5. converts executable relative `scripts/...` paths to checkout-absolute paths

### Claude Dependencies

The wrapper does not depend on a specific Pi orchestration or web-search extension. I love this about pi that everything is fully self configurable.

When ARS requests multi-agent execution, Pi is instructed to inspect available tools and installed skills for:

- subagent support
- workflow orchestration
- parallel-agent execution

If none is available, roles run sequentially in the current agent and the degraded execution mode must be disclosed.

When ARS requests `WebSearch`, `WebFetch`, or `/websearch`, Pi is instructed to locate an installed web-search or retrieval capability. If none exists, it must report the limitation and must not claim that sources were verified.

### Extra Command / Diagnostics

The current prototype wrapper adds:

```text
/ars-pi-doctor
```

This command runs without an LLM call and reports:

- discovered orchestration tools or skills
- discovered web-search or retrieval capabilities
- Python availability
- PyYAML availability
- Pandoc availability
- tectonic availability
- sandbox status
- Claude-hook compatibility status

The wrapper reports missing optional dependencies but does not install them.

## Known differences from Claude Code

This is a methodology compatibility wrapper, not full Claude Code runtime parity.

1. `hooks/hooks.json` is not executed by Pi.
2. The Claude `PreToolUse` write-scope guard is therefore not deterministically enforced.
3. Claude `SessionStart` hooks and plugin update notices are unavailable.
4. `model: sonnet` command frontmatter is ignored; Pi inherits the active model.
5. The 39-agent ensemble is only isolated or parallel when the user's Pi installation provides a compatible orchestration capability.
6. Without orchestration support, sequential fallback cannot provide independent context windows.
7. Web retrieval depends on user-installed Pi capabilities.
8. Sandbox rules may require explicit read access to the ARS checkout.

The wrapper will not claim full Claude parity or claim multi-agent execution when only sequential fallback occurred.

HannesOberreiter:
## Summary

Adds a community-maintained Pi wrapper for the four original ARS skills and all
16 `/ars-*` commands.

The wrapper keeps upstream ARS files authoritative, supports direct Git
installation, scopes compatibility instructions to ARS sessions, and provides
capability diagnostics plus degraded-mode disclosure.

Design approved in https://github.com/Imbad0202/academic-research-skills/issues/631

## Eval impact

No eval impact.

## Platform port?

Yes. This adds the new `pi/` platform port approved in:
https://github.com/Imbad0202/academic-research-skills/issues/631

## Full-run workflow stats

I did the example full run, this are the stats:

| Workflow | Input | Output | Cache read | Total | Runtime | Cost |
|---|---:|---:|---:|---:|---:|---:|
| Main run | 866,369 | 60,291 | 39,141,888 | 40,068,548 | 12h 35m | $36.02 |
| Finish run | 585,807 | 34,521 | 9,486,336 | 10,106,664 | 34m 25s | $10.68 |
| **Combined** | **1,452,176** | **94,812** | **48,628,224** | **50,175,212** | **13h 9m** | **$46.70** |

## Checklist

## Validation evidence

[`examples/pi/`](examples/pi/) contains a privacy-minimized completed `/ars-full`
run: representative output from every stage, degraded-mode disclosures,
Stage 5 provenance refusal/PASS, terminal state, inventory, and checksums.

Raw provider/session logs, internal identifiers, source downloads, redundant
binaries, and build caches are intentionally excluded.

Closes #631


Imbad0202 (changes requested):
Cross-model review (gpt-5.6-sol xhigh plus an independent security pass): one blocking finding, two advisories. The inert root manifest is accepted as-is.

1. **[P1] Model-invoked skills bypass activation.** Pi 0.83.0 exposes skill descriptions in the system prompt for automatic model invocation (docs/skills.md), and the four ARS skills carry no `disable-model-invocation` frontmatter. An ordinary prompt like "write an academic paper" can auto-invoke an ARS skill while `arsActive` stays false, so the compatibility note, path guidance, and degraded-mode disclosure never inject. That also undercuts the "System-prompt scope" section in pi/README.md and the `ordinary_prompt_ars_text_active: false` evidence field. Since the wrapper must not modify upstream SKILL.md files, either inject the note unconditionally (its first line already self-scopes to ARS usage) or hook whatever fires on model skill invocation, and document any residual gap.
2. **[P2]** Script-path rewrite runs after argument substitution (wrapper.js:70-76), so user args containing text like `python scripts/foo.py` get mutated. Rewrite the command body before inserting args.
3. **[P2]** `/tree` navigation leaves `arsActive` stale: state recovers only on `session_start`, but `session_tree` changes the branch without one. Recompute from `getBranch()` there too.

Security pass on wrapper.js and the evidence bundle: no findings. For closes-check, add `Closes #631` to the PR body.

Happy to merge once the P1 lands; the P2s can ride the same push or a follow-up.

HannesOberreiter:
## P1

While ARS mode is inactive, the wrapper **now hides this package’s skill entries** from ordinary system prompts, **preventing automatic model invocation** without modifying upstream `SKILL.md` files. I think this is the best solution as I don't like to have always a extended system prompt only for the ARS skills.

Explicit `/ars-*` and direct `/skill:*` requests still activate compatibility before the same agent run. `/ars-pi-start` enables automatic selection, and `/ars-pi-stop` disables it again.

## P2

- Script-path rewriting now runs on the command body before argument substitution, so user-provided arguments remain unchanged.
- The `session_tree` event now restores `arsActive` from the selected branch.

Imbad0202 (changes requested):
Thanks for the fast turnaround. I verified the delta (`790de95..906ed2e`) change by change, and everything lands where it should:

- **Script-path rewrite before argument substitution: VERIFIED.** The rewrite now runs on the command body only, and the `python scripts/user-provided.py`-as-argument vector is closed. The function-replacer form also avoids replacement-token expansion. Good fix.
- **`session_tree` state restoration: VERIFIED.** Pi switches the leaf before emitting `session_tree`, so recomputing from `getBranch()` is the right recovery point. I found no stale-closure or cross-branch leak.
- **The evidence bundle kept pace with the changes**, which I want to acknowledge explicitly: the new `system_prompt_scope` fields, the colliding-skill preservation probe, and the disclosure that the post-run changes did not get another full pipeline execution are all the right instincts. All 23 SHA256 entries verify here, and `node --test pi/wrapper.test.mjs` passes 6/6.

On the P1 itself (hide-while-inactive), the mechanism is correct for the idle, canonical-path case — but a second review round found two ways it fails outside that case. As in the first round, I ran a cross-model review pass over the delta and then confirmed the findings myself against the `pi-mono` v0.83.0 source and local reproductions before writing this.

## New findings

**[P1] Symlink-spelled load paths defeat the hiding.**
`pi/wrapper.js:5` derives `repoRoot` from `import.meta.url`, which under default Node loading (no `--preserve-symlinks`) is the canonical realpath spelling. Pi, on the other side, lexically resolves each resource path and computes a canonical path only for deduplication — the lexical spelling is what stays in the loaded path (`resource-loader.ts` `mergePaths`, v0.83.0). So when Pi retains a symlink spelling in a skill's `<location>`, the two sides diverge, every `skillLocations` membership test misses, and all four ARS skills stay visible to ordinary prompts. macOS makes this easy to hit: `/tmp` is a symlink to `/private/tmp`, and I reproduced exactly that split locally — the wrapper resolves itself to `/private/tmp/...` while a block carrying the lexical `/tmp/...` location survives `hideArsSkills` untouched. A local install from a symlinked checkout is a realistic way to land in this state.
Fix direction: XML-decode the location value, then compare canonical filesystem identity (realpath on both sides, with a guard for paths that no longer exist) instead of raw strings.

**[P1] Streaming and RPC ordering bypass activation, so the "same agent run" guarantee in `pi/README.md` does not hold unconditionally.**
Verified against v0.83.0 `agent-session.ts` / `rpc-mode.ts`:

1. For a file-backed command like `/ars-plan` arriving while the agent is already streaming: `AgentSession.prompt()` runs the `input` handlers (your transform fires, `arsActive` flips), but the expanded text is then queued via steer/follow-up and the method returns — `before_agent_start` only fires on the non-streaming path, and `_systemPromptOverride` is cleared only when the whole run ends (`_runAgentPrompt`'s `finally`). The skill content is delivered into a run whose system prompt still hides the skills and carries no wrapper-injected compatibility note or disclosure instruction.
2. `/ars-pi-stop` is a registered extension command, so it executes immediately even mid-stream — but the same override lifetime means the in-flight run keeps its skills-visible prompt until it ends.
3. The RPC `steer` / `follow_up` commands (`rpc-mode.ts:417,422`) call `session.steer()` / `session.followUp()` directly, and those expand `/skill:*` themselves without emitting `input` at all — the wrapper never sees the invocation, so a directly-steered `/skill:academic-pipeline` runs with `arsActive` false and no wrapper-injected compatibility layer. `steer` and `follow_up` are first-class commands in Pi's supported RPC surface, so I consider this path in scope for the port.

Whichever way you take this, the README claim needs narrowing: scope §System-prompt scope (and the corresponding evidence fields) to the idle-prompt path and state the mid-stream/RPC residual explicitly — that is consistent with the original ask ("document any residual gap"). Optionally, in addition, close what code can close (e.g., when activation happens mid-stream, also queue the compatibility note as a message for the current run — noting that a queued message is guidance to the model, not a system-prompt override, so the narrowed wording is needed either way).

## Advisories (can ride the same push or a follow-up)

- **[P2] `$ARGUMENTS`/`$@` two-pass substitution still mutates user arguments.** The second `replaceAll("$@", ...)` re-scans text the first pass just inserted: an argument `alpha$@omega` becomes `alphaalpha$@omegaomega` (reproduced locally). A single-pass `/\$ARGUMENTS|\$@/g` function replacement closes it. This is a pre-existing residual, not something the delta introduced.
- **[P2] `hideArsSkills()` can cross block boundaries on alternate skill-block formatting.** A `</skill>` on the same line as content lets the match run into the next block, and the deletion decision is then keyed to the first `<location>` in the over-matched span — I reproduced an adjacent non-ARS block being deleted along with the ARS one. The multiline layout your current probes exercise is handled correctly, hence advisory; a match that refuses to cross the next `<skill>` boundary removes the class. Minor related nit: the test file builds `new RegExp(location)` from unescaped absolute paths, which can fail or silently weaken assertions on checkouts whose path contains regex metacharacters.

## Process note for this and future rounds

The pattern behind both P1s: the README's claims are unconditional, while the evidence bundle's probes cover the idle-prompt, canonical-path case — no streaming or steer/follow-up scenario appears in the bundle. So for the load-bearing system-prompt claims in this PR, the ask is: contributor-run evidence covering the claim's stated scope, or the claim narrowed to what the evidence covers. I intend to codify that claims-evidence bar in the platform-ports section of CONTRIBUTING so it applies to every port rather than being invented per-review. Per that same policy, Pi runtime expertise stays with you and Pi-specific issues keep routing to you; what I review on my side is the alignment, plus ARS's own principles surviving the port — degraded-mode disclosure and human-in-the-loop above all. This round I did go into the Pi source to confirm the findings before publishing them; I'm saying so explicitly so the division of labor is stated rather than discovered.

## Mechanics

- If #631 is the issue this PR resolves — I believe it is — add `Closes #631` to the PR body; the closes-check gate needs an accepted issue reference (or its documented override) to go green.
- After your next push, I'll approve any CI runs GitHub holds for approval so the gates actually execute.

Happy to do the next round as soon as the two P1s land or are scoped. The port itself remains work I want in the tree — the evidence discipline you've shown across these rounds is exactly what makes a community-maintained port viable.

HannesOberreiter:
- `5a9c80b`: scoped system-prompt guarantees to idle prompts and documented the mid-stream/RPC residuals.
- `5175c44`: XML-decode and canonicalize skill paths before comparison, covering symlink-spelled checkouts and missing paths.
- `b66caa3`: changed argument expansion to single-pass substitution.
- `ec7dc2e`: bounded skill-block matching so alternate formatting cannot remove adjacent skills.

Validation:

- `node --test pi/wrapper.test.mjs`: 9/9 PASS
- Evidence checksums: 23/23 PASS
- Independent Pi 0.83.0 runtime smoke test: 5/5 PASS using real commands/SKILL files and captured provider payloads
- Working tree clean

The smoke test verified inactive hiding, `/ars-plan` and direct `/skill:academic-pipeline` activation/injection, `/ars-pi-stop`, and colliding-skill preservation. No full pipeline rerun was performed; the evidence bundle continues to disclose that boundary.

Also added `Closes #631` to the PR body.

Imbad0202 (approved):
Round-3 verification of the delta (`906ed2e..ec7dc2e`), same two-track process as the earlier rounds: a cross-model pass over the delta, findings confirmed first-hand before writing this.

- **Symlink/XML-escaped location hiding: VERIFIED.** Canonical filesystem identity is compared on both sides, missing paths fail toward preserving the block, and the new filesystem-backed symlink test fails on the pre-fix comparison, so it witnesses the fix rather than assuming it.
- **Idle-scope documentation: VERIFIED.** The narrowed §System-prompt scope and the evidence bundle now claim exactly what the probes cover; the mid-stream and `steer`/`follow_up` residuals are stated rather than implied away. This is the claims-evidence bar from last round, met.
- **Single-pass substitution: VERIFIED**, with the `alpha$@omega` regression pinned.
- **Bounded block matching: VERIFIED.** The match refuses to cross the next `<skill>` opener, inline closing tags are handled, and absent closures fail toward preserving content.

Local runs here: 9/9 tests, 23/23 checksums. CI is green; the closes-check failure you may have seen was my side (the held run predated your body edit and validated a stale payload), fixed by re-triggering.

One advisory, no action needed for this PR: `decodeXml` covers the five named entities only. If Pi ever emits numeric character references, the failure direction is skill-stays-visible, which is the safe side; noting it so the boundary is on record.

Approving. Three rounds at this evidence discipline is exactly what makes a community-maintained port viable; thank you.

Issue #128, opened 2026-05-17
v3.9.1: housekeeping — extract shared client utilities + dedup resolvers


Imbad0202:
## Scope

v3.9.0 (#102, merged) shipped intentionally mirroring `scripts/semantic_scholar_client.py` structure across 3 clients (S2, OpenAlex, Crossref) and 2 migration tools (v3.7.3, v3.9.0). The mirror pattern was a design decision for code locality, but `/simplify` agent review identified extractable duplication that's worth resolving now that the 3-client family is stable.

This issue tracks **non-blocking housekeeping** — no behavior change, just dedup + helper extraction. None of these are urgent; they prevent drift between sibling clients.

### 1. Extract `_text_similarity.py` shared utility (3-way dedup)

**Duplicated across:** `scripts/semantic_scholar_client.py`, `scripts/openalex_client.py`, `scripts/crossref_client.py`

**What's identical:**
- `_PUNCT_TRANSLATION = str.maketrans({c: " " for c in string.punctuation})` constant
- `_normalize_title(s: str) -> str` function (lowercase + punctuation strip + whitespace collapse)
- `_similarity(a, b) -> float` using `difflib.SequenceMatcher`
- `_TITLE_SIMILARITY_THRESHOLD = 0.70` magic number (PaperOrchestra precedent)

**Suggested refactor:** Create `scripts/_text_similarity.py` with the 3 helpers + 1 constant. Import from each client. ~4-line module + 3-line import each side.

### 2. Extract shared API client retry / backoff constants

**Duplicated across:** 3 clients

**What's identical:**
- `_BACKOFF_SECONDS = 2.0`
- `_MAX_RETRIES = 3`

Either add to `scripts/_text_similarity.py` or new `scripts/_api_client_constants.py`. Lower priority than #1 — but adding alongside has no extra cost.

### 3. Deduplicate `resolve_openalex_unmatched` / `resolve_crossref_unmatched`

**File:** `scripts/contamination_signals.py`

**What's near-identical:** Two resolvers differ only in client type + exception type. Both call `client.doi_lookup_with_title_check(doi, title)` + `client.title_search(title)` with identical control flow.

**Suggested refactor:** Private `_resolve_by_doi_then_title(entry, client) -> bool | None` helper. Both named resolvers become 1-line wrappers. Keep both named API surface for spec sections §3.4 + §3.5 — only consolidate the body.

### 4. Parallelize OpenAlex + Crossref calls per entry in migration tool

[...]

Issue #431, opened 2026-06-13
Title-match criterion (SequenceMatcher ≥0.70) accepts distinct works as matched — harden scoring


Imbad0202:
Split out from #250 as the production-code defect that the strict-xfail pin in #430 documents. #250 is now re-scoped to "resolver-client-layer test exists"; this issue tracks the actual fix.

## The defect

The title-fallback match criterion is `difflib.SequenceMatcher.ratio() >= 0.70` over punctuation-stripped titles (`scripts/_text_similarity.py` `_similarity` / `_TITLE_SIMILARITY_THRESHOLD`). Character-level ratio scores **distinct works** far above 0.70 whenever their titles share a long substring, so the resolver `title_search` accepts a wrong record as `matched`:

| Cited work | Surfaced (different) work | ratio | current verdict |
|---|---|---|---|
| `Deep Residual Learning for Image Recognition` | `Deep Residual Learning for Image Recognition on Embedded Devices` | 0.815 | accepted ❌ |
| `Attention Is All You Need` | `Attention Is Not All You Need` | 0.926 | accepted ❌ |
| `A Survey of Reinforcement Learning in Healthcare` | `A Survey of Deep Reinforcement Learning in Robotics` | 0.808 | accepted ❌ |
| `Federated Learning for Mobile Keyboard Prediction` | `Federated Learning for Mobile Health Monitoring` | 0.729 | accepted ❌ |

`_resolve_doi_then_title` (`scripts/contamination_signals.py`) runs the title fallback with **no year/author re-check** to catch this — a passing candidate is collapsed straight to `matched`. The same criterion is shared by the Crossref / OpenAlex / Semantic Scholar clients via `_text_similarity`, so the defect is cross-resolver, not Crossref-specific.

## Impact

This is the false-positive path #250 named: a real-but-unindexed citation whose title search surfaces a near-identically-titled *different* paper is reported as verified (`matched` ⇒ `lookup_verified: true`). It only fires on the title-fallback path (no DOI hit / DOI miss), so DOI-keyed citations are unaffected — hence p2, not p1.

## Existing pin

`scripts/test_title_fuzzy_false_positive_xfail.py` (#430) holds two `xfail(strict=True)` tests asserting the correct behavior (distinct work rejected). When this issue is fixed they flip to a build-failing **xpass**, forcing removal of the marker and conversion to real passing assertions. That is the acceptance signal for this issue.

## Candidate fixes (design TBD)

[...]

PR #799, opened 2026-08-26
fix: apply Chinese-aware title matching in the four index resolvers (#798)

### LINKED ISSUE #798: CJK titles fail the shared exact-title gate in the four index resolvers (real Chinese papers render as DOI_MISMATCH / unresolvable)
The four index resolvers (Semantic Scholar / OpenAlex / Crossref / arXiv) read the ASCII-centric helpers in `scripts/_text_similarity.py`. `chinese_literature_client.py` already carries a Chinese-aware `normalize_cn_title` / `has_cjk` (added with the #431 §"Chinese title matching" work), but the four resolvers never call it.

Consequence: a Chinese paper that an index serves in a different-but-legitimate typesetting fails to match, and reduces to `unresolvable` / `DOI_MISMATCH` — the same state a fabricated citation produces.

## Reproduction

No network required, from repo root:

```python
import sys; sys.path.insert(0, 'scripts')
from _text_similarity import exact_normalized_title, _similarity

cited   = "基于ProEXC的宫颈癌筛查研究"          # as cited by the author
indexed = "基于ＰｒｏＥＸＣ的宫颈癌筛查研究。"   # as an index serves it (fullwidth Latin + 。)

exact_normalized_title(cited, indexed)   # -> False
_similarity(cited, indexed)              # -> 0.606   (floor is 0.70)
```

Measured across the variants recorded in `deep-research/references/chinese_literature_api_protocol.md`:

| variant of one identical title | shared `exact_normalized_title` | shared ratio |
|---|---|---|
| identical string | ✅ True | 1.000 |
| fullwidth Latin (`ＰｒｏＥＸＣ`) | ❌ False | 0.625 |
| trailing `。` | ❌ False | 0.970 |
| interior spaces around Latin runs | ❌ False | 0.941 |
| `《》` title wrapper | ❌ False | 0.941 |
| ideographic space U+3000 | ❌ False | — |

## Root cause

Two ASCII assumptions in `_normalize_title`:

- **`.lower()` folds case but never width.** `Ｐ` is U+FF30, `P` is U+0050. Nothing maps between them, so `ProEXC` and `ＰｒｏＥＸＣ` normalize to different strings. Chinese typesetting routinely renders embedded Latin fullwidth so it aligns to the Han character grid, so both spellings of one title legitimately exist across indexes.
- **`string.punctuation` is ASCII-only.** It contains none of `。`, `《》`, `「」`, or U+3000, so they survive normalization and break equality.

## Why the fuzzy fallback does not save it

The 0.70 ratio should catch near-misses. It cannot, for a structural reason: a Han character is a whole word, so Chinese titles are very short in codepoints. The same six-codepoint corruption is **37.5%** of a 16-character Chinese title but **8%** of its 68-character English equivalent:

```
identical pair, Chinese title  : 0.625   (fails)
identical pair, English-length : 0.912   (passes)
```

A threshold tuned on English titles is the wrong threshold for a script with roughly 4× the information density per character. This is why the defect is invisible to the existing English test corpus.

Worse, the ratio is **anti-correlated** on the motivating pair:

```
identical title (fullwidth variant) : 0.606
a genuinely DIFFERENT paper         : 0.645   <- ranks higher
```

## Both affected paths

1. **DOI-keyed cross-check** — e.g. `crossref_client.doi_lookup_with_title_check:187`, mirrored in the other three. Gates on the ratio **alone**. 0.625 < 0.70 → returns `None` → `DOI_MISMATCH`. A correct DOI resolving to the correct paper is reported as a mismatch.
2. **Title-fallback search** — e.g. `crossref_client.title_search:214-218`. Requires ratio ≥ 0.70 **and** exact-normalized equality. Fails both → candidate discarded → `unresolvable`.

## Impact

Both paths feed the `*_unmatched` contamination signals, which increment *k* in the v3.9.0 triangulation matrix. A genuine Chinese paper missing across enough indexes can render as `CONTAMINATED-TRIANGULATION-UNMATCHED` — the strongest fabrication signal the suite emits. `chinese_literature_api_protocol.md` names this outcome directly: a real paper filed "at **P0**, next to the word 'fabricated'."

Per that same protocol doc, OpenAlex indexes roughly 37% of Chinese core-list journals, so the affected papers *are* present in the indexes being queried — they are simply not being recognized.

## Proposed fix

Promote `has_cjk` / `normalize_cn_title` from `chinese_literature_client.py` into `scripts/_text_similarity.py` so all four resolvers apply the same Chinese-aware rule, keeping the existing conservatism (no Simplified/Traditional folding, no whole-string NFKC, no broad punctuation deletion, fuzzy ratio still excluded from `_cn_titles_match`). Gate the new comparison on **both** sides carrying a Han ideograph so every non-CJK verdict is provably unchanged.

Happy to send a PR — I have the fix and tests ready.

akshath-raj:
Closes #798

Makes the four index resolvers (Semantic Scholar / OpenAlex / Crossref / arXiv) apply the Chinese-aware title rule that `chinese_literature_client.py` already had. Before this, a real Chinese paper served by an index in a different-but-legitimate typesetting reduced to `unresolvable` / `DOI_MISMATCH` — the same state a fabricated citation produces.

This is wiring, not invention: the algorithm and its conservatism are the maintainer's own, promoted from where they already lived. The promotion is behaviorally equivalent, not byte-identical: it hoists locals to module constants, precompiles the regex, and rewrites comments — equivalence of behavior is what the tests pin.

## Change

**One production file.** `scripts/_text_similarity.py` gains `has_cjk` / `normalize_cn_title`; `chinese_literature_client.py` re-imports them rather than keeping a private copy (the #128 anti-drift goal — the two copies could otherwise diverge again).

- `exact_normalized_title` gains the Chinese-aware form as an **additive third branch**, alongside the base and dotted-acronym forms.
- `_similarity` folds it in through the **existing `max`**. This is what repairs the DOI path, which gates on the ratio alone — so the four client files need no changes at all.

Both are gated on **both** sides carrying a Han ideograph.

## Why `_similarity` returns 1.0 on that branch

Not a sentinel. Under the Chinese-aware normalization the two strings are byte-equal, so their `SequenceMatcher` ratio genuinely *is* 1.0 — the same shape as the existing acronym pre-pass, which already makes `_similarity` a max over equivalent normalization forms rather than a single ratio.

The CJK form is folded in **only when it makes the titles equal**. An unrelated CJK pair is never re-scored, so Han-overlap noise is not inflated.

## Preserved

- Simplified/Traditional is still not folded (lossy for proper nouns).
- No whole-string NFKC/casefold (it would collapse `2²`/`22`, `Straße`/`Strasse`).
- No category-wide punctuation deletion (it would collapse `ER+`/`ER-`, `4.5%`/`45%`).
- Cross-script and romanized pairs still cannot match — there is no translation oracle, so a difference stays a non-match rather than becoming chimeric-citation evidence.
- An empty normalized key never matches (`《》` and `「」` both normalize to `""`).
- The fuzzy ratio remains excluded from `_cn_titles_match` itself. That decision is untouched.

## Evidence

Measured on the pair recorded in `chinese_literature_api_protocol.md`:

| | pre-fix | post-fix |
|---|---|---|
| identical title (fullwidth variant) | 0.606 | **1.000** |
| genuinely different paper | 0.645 | 0.645 (unchanged, verified to 4 dp) |

The pre-fix ratio was **anti-correlated** — the wrong paper outranked the correct one. The repair lifts only the true match; the unrelated score is untouched. Pinned by `test_true_match_outranks_an_unrelated_paper`.

Under the Chinese-aware form the ratio also regains discriminative power on this pair (1.000 vs 0.510 on the protocol doc's own comparison pair, where the base form separated only 0.566 from 0.510).

## Tests — 32 new

Every one was verified to fail against the pre-fix module, including the five integration tests (written after the fix, so proven by reverting `_text_similarity.py` and re-running).

- **Non-destructiveness is proven, not sampled.** `_pre_fix_exact_normalized_title` restates the pre-fix formula in full, and the repair is asserted to agree with it on every non-CJK pair — case, ASCII punctuation, dotted acronyms, the `D. H.` base-form carve-out, `A/B`, `R&D`, distinct related works, and the empty/whitespace degenerate cases. A companion `_pre_fix_similarity` oracle restates the entire pre-fix `_similarity` — base ratio **plus** the dotted-acronym branch — and off the CJK path the repair is asserted to reproduce it with `assertEqual`, so a *raised* non-CJK score fails as loudly as a lowered one.
- **5 integration tests through the real `CrossrefClient`**, covering both broken paths plus the negative guards (different CJK paper, cross-script shadow title).
- **Resolver reach is pinned**: each of the four clients is asserted to resolve the *same* helper object, so a future re-implementation cannot silently drift.
- Unit coverage for the normalizer, `has_cjk` (including that fullwidth Latin alone is *not* CJK), Simplified/Traditional non-folding, and scientific-name guards.

Full suite (`PYTHONPATH=. GITHUB_DEFAULT_BRANCH=main pytest scripts/ tests/`): **9256 passed, 3 skipped**, plus the pre-existing failure noted below. Collection count 9228 → 9260 on clean `main`, so the +32 is exact.

## Please look at these deliberately

**1. I modified an existing test.** `test_legitimate_variants_match_despite_a_sub_threshold_fuzzy_ratio` asserted `_similarity(...) < 0.70` as its premise — and this change makes that false by design. I rewrote the premise to measure the **base-form** ratio (the thing that was actually dangerous, and the number the docstring was really about), kept both `_cn_titles_match` assertions — now run inside a `monkeypatch.context()` that makes any read of `_similarity` raise, so the match is *forced* rather than observed while the ratio happened to agree — and added the new invariant that the shared helper no longer contradicts the Chinese-aware rule. The test's purpose is intact, but the hunk deserves a real read.

**2. Out of scope, deliberately left alone.** The base ASCII normalization already collapses `ER+`/`ER-` and `p53`/`P53` (it maps ASCII punctuation to whitespace and lowercases), and `exact_normalized_title` ORs that form in — so this additive change neither causes nor removes that. Relatedly, the 0.645 an *unrelated* CJK pair scores is a pre-existing weakness of the fuzzy floor for CJK generally; untouched here. Both are noted rather than fixed, to keep the diff to one concern.

**3. Pre-existing unrelated failure.** `scripts/test_evidence_rows.py::test_gfm_bare_urls_emails_and_schemes_cannot_autolink` fails on clean `main` as well (verified by stashing). Not touched.

## Docs

`chinese_literature_api_protocol.md` §"Chinese title matching" gains a "Repaired" note; its measurement table is preserved as the historical record and its tense corrected. CHANGELOG entry added under `[Unreleased]`.


Imbad0202 (changes requested):
Thanks for this. The issue is one of the best-evidenced reports this repo has received, and the fix takes the right shape: promoting the existing #431 rule rather than inventing a parallel one.

**One design point I checked and am deliberately leaving as-is.** Pairs like `《围城》` / `围城` and `研究...` / `研究` now normalize to the same key on all four resolvers. That widening is intentional and consistent with the documented #431 design in `chinese_literature_api_protocol.md` (conventional inert outer wrappers and terminal marks are typesetting noise). `_cn_titles_match` has applied the same equivalences on the Chinese DOI path since that work landed, and I verified the same pairs already return True on main's `chinese_literature_client`. Recording this here so the decision stays traceable.

**Requested changes (3):**

1. `test_legitimate_variants_match_despite_a_sub_threshold_fuzzy_ratio` no longer proves `_cn_titles_match` is ratio-independent: the pair it feeds now scores 1.0 under the repaired `_similarity`, so a regression that quietly made `_cn_titles_match` consult the ratio would still pass. Please force the dependency to fail (for example, monkeypatch `_similarity` to raise or return 0.0) and assert the match still holds.

2. `test_ratio_never_lowered_off_the_cjk_path` asserts `>=` against the base ratio only. A regression that raised a non-CJK score (even from 0.6 to 1.0) would pass. Please reconstruct the full pre-fix `_similarity` (including the dotted-acronym branch) the same way your `_pre_fix_exact_normalized_title` oracle does, and assert exact equality off the CJK path.

3. "Byte-identical" (PR body, CHANGELOG, `normalize_cn_title` docstring) overstates: the promotion hoists locals to module constants, precompiles the regex, and rewrites comments. Behaviorally equivalent is the claim your tests actually pin, so please say that.

**Optional, fine as a follow-up issue.** The wrapper-strip check is positional only, so a title whose first and last characters are marks from two different pairs gets mangled: `《红楼梦》与《金瓶梅》` normalizes to `红楼梦》与《金瓶梅`. Matching still works because both sides mangle identically, and the behavior is inherited from the original implementation, so I don't hold this PR to it. If you want to fix it here, a pairedness check before stripping would do; otherwise I'll open a follow-up.

I've approved the CI run. Once the three items land, this merges.

akshath-raj:
All three items landed in `86670f9`, pushed to the branch.

**1. Ratio-independence is now forced, not observed.** `test_legitimate_variants_match_despite_a_sub_threshold_fuzzy_ratio` installs a `_forbid_similarity` detonator that monkeypatches `_similarity` to raise on both binding paths — the `_text_similarity` module attribute (catching a qualified `ts._similarity(...)` call or a lazy in-function import) and `chinese_literature_client`'s own namespace (catching a module-level `from ... import _similarity`, already bound at import time and blind to the first patch). Inside that context, `_cn_titles_match(fullwidth, canonical)` and the reverse must still return `True`. A regression that ANDs the ratio back in now fails loudly instead of sailing past an agreement that happened by coincidence. The sibling `test_cn_titles_match_never_consults_the_fuzzy_ratio` reuses the same detonator to pin the negative half (the unrelated paper stays rejected without the ratio existing at all), so the invariant can't be satisfied by a helper that has simply stopped discriminating.

Mutation-verified: ANDing the ratio back in via a module-level `from`-import — the binding style the client does *not* currently use, i.e. the one only the second patch catches — trips both tests.

**2. Non-CJK ratios are now exact-equality against the full pre-fix formula.** Renamed to `test_ratio_unchanged_off_the_cjk_path`; the new `_pre_fix_similarity` oracle restates the entire pre-fix `_similarity` in full — base ratio plus the dotted-acronym branch, written out rather than captured so it can't drift with the module it checks — and `_NON_CJK_PAIRS` spans every branch (case, ASCII punctuation, dotted acronyms, the `D. H.` carve-out, `A/B`, `R&D`, degenerate empty cases). Asserted with `assertEqual` in both directions: a raised non-CJK score (0.6 → 1.0) now fails exactly as loudly as a lowered one.

Mutation-verified three ways: a raised base-branch score, the acronym branch deleted, and the acronym branch score halved — the old `>=` assertion caught none of them.

**3. Wording corrected on all three surfaces.**

- PR body: "promoted byte-identical" → "behaviorally equivalent, not byte-identical: it hoists locals to module constants, precompiles the regex, and rewrites comments — equivalence of behavior is what the tests pin." (The evidence table's "byte-identical" cell referred to the 0.645 score *value*, not the promotion; reworded to "unchanged, verified to 4 dp" to remove the collision.)
- CHANGELOG `[Unreleased]`: same correction.
- `normalize_cn_title` docstring: "Not byte-identical: … Behaviorally equivalent … Equivalence of *behavior* is what the tests pin."

I also refreshed the PR body's Tests section, which had gone stale against the new commit: 31 → 32 new tests, the "never lowered" phrasing replaced with the exact-equality description, and the counts corrected to 9256 passed / collection 9228 → 9260.

**Verification.** Targeted suites: `test_text_similarity.py` + `test_chinese_literature_client.py` + `test_crossref_client.py` → 237 passed. Full suite (`PYTHONPATH=. GITHUB_DEFAULT_BRANCH=main pytest scripts/ tests/`): 9256 passed, 3 skipped. The one failure, `scripts/test_evidence_rows.py::test_gfm_bare_urls_emails_and_schemes_cannot_autolink`, is unchanged and also fails on clean `main` (re-verified in a detached worktree at the merge-base, not by stashing). Collection delta measured against that same clean worktree: +32 exactly.

**On the optional wrapper-strip item:** agreed it's real, but it's inherited from the original implementation and matching is unaffected since both sides mangle identically. Happy to take a pairedness-check fix as a follow-up issue if you open one, or I can send it separately — your call which keeps the history cleaner.

Thanks for the close read on items 1 and 2 — both tests were passing for the wrong reason, and neither would have caught the regression it named.

Imbad0202:
Re-verified independently on my side: the detonator trips under a forced-regression mutation on the qualified-call path, the `_pre_fix_similarity` oracle agrees with the real pre-fix `_similarity` on a spot-check against clean main, and the targeted suites pass locally (215 + subtests). CI is running on `86670f9` now; I'll merge once it's green.

For the wrapper pairedness check I've opened #800, so this PR stays scoped to the resolver fix. It's yours if you want it, no obligation.

One datapoint I can't reproduce: `test_gfm_bare_urls_emails_and_schemes_cannot_autolink` passes on clean main here (markdown-it-py 4.0.0, linkify-it-py 2.1.0; the escaped forms produce zero `link_open` tokens). Could you share the markdown-it-py and linkify-it-py versions in the environment where it fails? That test's tail rides on third-party linkify behavior, and if a version combination flips it we should pin or guard it on our side, as a separate issue.

Imbad0202 (approved):
All three requested changes verified independently (detonator mutation-tested, oracle spot-checked against clean main's pre-fix implementation, wording corrected on all three surfaces). CI green including the full pytest suite. Merging.

akshath-raj:
Versions in the environment where it fails: markdown-it-py **2.2.0**, linkify-it-py **2.0.3** (Python 3.13.9) — so two majors behind your 4.0.0 / one patch behind your 2.1.0.

The failure does reproduce on clean `main` here — re-verified just now in a detached worktree at the merge-base (`127ff85`, pre-PR): under 2.2.0's linkify pass, the escaped/bare forms in that test produce `link_open` tokens, which is what trips the `assert all(token.type != "link_open" ...)`. Nothing about the failure touches this PR's change.

Agreed it rides entirely on third-party linkify behavior across major versions, and that pinning/guarding on our side (e.g. a minimum-version floor or a version-aware expectation) is the right call as a separate issue.

PR #804, opened 2026-08-27
fix: strip CJK outer wrapper marks only when they enclose one balanced unit (#800)

### LINKED ISSUE #800: normalize_cn_title strips outer wrapper marks positionally without verifying the pair matches (#798/#799 follow-up)
## Context

Surfaced during the #799 review (follow-up to #798). Inherited from the original `chinese_literature_client.py` implementation (#431); the #799 promotion moved the behavior into `scripts/_text_similarity.py` unchanged, and this issue tracks the pre-existing weakness rather than anything #799 introduced.

## Behavior

`normalize_cn_title` strips wrapper marks (`《》「」『』（）()"″`, etc.) by *position* (leading/trailing) without verifying the stripped pair actually matches:

```
《红楼梦》与《金瓶梅》  →  红楼梦》与《金瓶梅
```

The leading `《` and the trailing `》` are removed even though they belong to two different bracket pairs, leaving interior orphaned marks.

## Why this is not urgent

Both sides of every comparison pass through the same normalization, so a mangled key still matches its identically-mangled counterpart. Matching correctness is unaffected today, and we found no exploitable asymmetry during the #799 security pass.

## Why it is still worth fixing

- The normalized key is a semantic anomaly waiting for a future consumer that uses it single-sided (display, logging, cache keys, or a resolver that compares against an externally-normalized key).
- A pairedness check (strip the outer pair only when the leading and trailing marks form a matching pair) is small and local.

## Acceptance sketch

- Strip an outer wrapper only when the first and last characters form a matching pair from the wrapper table.
- Titles like `《红楼梦》与《金瓶梅》` keep their marks (the outer pair does not enclose the whole title as a unit; conservative behavior is acceptable here as long as both callers stay consistent).
- Both consumers (`chinese_literature_client` and the four index resolvers via `_text_similarity`) must change together; the #799 oracle tests should be extended to pin the new behavior on both.

@akshath-raj offered in #799 to take this as a follow-up; it's theirs if they want it.

akshath-raj:
Closes #800

Follow-up to #798/#799, as offered there. The wrapper strip inherited from #431 was positional: it removed the first and last characters whenever they matched as a wrapper pair *type*, without checking they belonged to the same bracket pair. `《红楼梦》与《金瓶梅》` normalized to `红楼梦》与《金瓶梅`, leaving an orphaned `》` mid-key.

Matching correctness was never affected — both sides of every comparison pass through the same normalization — so this is the semantic-anomaly cleanup the issue describes, not a correctness repair.

## One thing worth flagging about the acceptance sketch

The issue says "strip an outer wrapper only when the first and last characters form a matching pair from the wrapper table." Implemented literally, that is a **no-op on the motivating case**: in `《红楼梦》与《金瓶梅》` the leading `《` and trailing `》` *do* form a matching pair from the table. They just belong to two different brackets.

So the check has to look at the interior. `_outer_pair_encloses` strips only when the outer marks form a pair **and** the interior is balanced within that pair's own family (see round 1 below — the first version scanned all six families, which was unsound). That catches both failure modes, which are genuinely distinct:

- **stray closer** — `《红楼梦》与《金瓶梅》` → pop from an empty stack
- **unclosed opener** — `“研究与“实践”` → stack non-empty at the end

Each has its own test. I mutation-verified the naive first/last-character version to be sure this wasn't over-engineering: it fails 4 tests, so the interior scan is load-bearing. (Round 1 of review narrowed *which* marks the scan looks at — see below.)

## Behavior

| input | output | |
|---|---|---|
| `《红楼梦》与《金瓶梅》` | unchanged | the bug |
| `“研究”与“实践”` | unchanged | stray closer, quote pair |
| `“研究与“实践”` | unchanged | unclosed opener |
| `《围城》` | `围城` | still strips |
| `《基于「ProEXC」的研究》` | `基于「ProEXC」的研究` | nested balanced interior still unwraps |
| `《》` / `「」` | `""` | empty-wrapper guard preserved |
| `《Alzheimer’s病中ＰｒｏＥＸＣ表达》` | `Alzheimer’s病中ProEXC表达` | apostrophe is not a wrapper (review P1) |
| `《「研究』》` | `「研究』` | other-family interior marks are content |

The #798 motivating ISTIC pair still matches, and `围城` ≡ `《围城》` still holds.

## Review round 1 — family-scoped interior scan

The first version scanned the interior against **all six** wrapper families. That was unsound, and @Imbad0202 caught it: `’` is the closer of `‘` but also the English apostrophe, so the lone `’` in `《Alzheimer’s病中ＰｒｏＥＸＣ表达》` looked like an unbalanced quote and vetoed a genuine `《…》` wrap. Verified against `main`: exact **True → False**, ratio **1.0 → 0.6818** — under the 0.70 floor, so both resolver paths failed at once. That is the failure class #798 repaired, not something #800's conservative-direction blessing covers.

The scan is now scoped to the outer pair's **own family**. This costs nothing the check was buying: any mark that can orphan the outer pair is by definition of that pair's own family. Every row above survives, and the two dead lookup maps are removed.

## One deliberate narrowing

`《红楼梦》与《金瓶梅》` no longer matches a pre-mangled `红楼梦》与《金瓶梅`. That is the conservative direction the issue blesses ("conservative behavior is acceptable here as long as both callers stay consistent"), and it is pinned with an explicit `assertFalse` in `test_mangled_key_still_matches_its_identically_mangled_counterpart` rather than left to be discovered. Flagging it because it is the one verdict this PR changes.

## Both consumers

The acceptance sketch requires `chinese_literature_client` and the four resolvers to change together. Since #799 the client re-imports the shared function, so identity already covers it — `test_pairedness_behavior_is_shared_by_the_cjk_client` pins it **behaviorally** as well, so a future private copy cannot hide the drift.

## Non-CJK — scope of the claim, stated precisely

The original wording ("non-CJK behavior is provably untouched") was broader than the code, per review advisory 3. Narrowed and pinned:

- **On the two `has_cjk`-gated paths** — `exact_normalized_title`'s third branch and `_similarity`'s CJK fold — every Han-free verdict and ratio is unchanged, asserted by the #799 oracles. `_NON_CJK_PAIRS` gains two mark-carrying Han-free entries (`《Attention Is All You Need》` vs the bare title) so a leak through the gate would fail loudly.
- **The client's `_cn_titles_match` is NOT gated** — it calls `normalize_cn_title` directly, so a mark-carrying Han-free title can change verdict there. Verified: `《Hamlet》and《Macbeth》` no longer matches a pre-mangled `Hamlet》and《Macbeth`. Accepted rather than gated, since that path is DOI-keyed and Chinese-corpus-only in practice; pinned by `test_ungated_client_path_narrowing_is_explicit`.
- **The empty-wrapper guarantee is a `_cjk_titles_match` property**, not a global one: `exact_normalized_title("《》", "《》")` is still True through the base branch, exactly as before this change. Pinned by `test_empty_wrapper_guard_is_a_cjk_branch_property`.

Your O’Connor example specifically now **matches again** — the apostrophe was the cause, so the P1 fix resolved it. The general hole you identified is real regardless, hence the narrowing above.

## Also in this round

- **Module docstring** (advisory 2): "behaviorally equivalent to the implementation this was promoted from" now carries a historical qualifier — true of the #798/#799 promotion, false since #800 deliberately changed the behavior.

## Verification

11 new tests. Every branch of the scan is mutation-verified:

| mutant | killed by |
|---|---|
| pre-#800 positional strip | 4 tests |
| naive first/last-pair check (the issue's literal sketch) | 4 tests |
| family-blind scan (the P1 regression) | `…apostrophe_in_embedded_english…` |
| clamp instead of refuse (depth-0 stray closer) | `…stray_interior_closer…` |
| drop the trailing `depth == 0` check | `…unclosed_interior_opener…` |

On advisory 1 — worth recording, since my first attempt at that test **did not discriminate**. Clamping absorbs one closer, so any title with equal opener/closer counts (every natural case, `《红楼梦》与《金瓶梅》` included) ends at depth 1 and is refused by the *other* branch anyway. Discriminating needs interior closers to outnumber openers by exactly the clamp count, which no natural title shape produces — so that test uses a documented synthetic asserted against the helper directly, and says why in its docstring so it is not "simplified" back later.

Full suite on this branch: **9270 passed, 4 skipped, 0 failed.**

Worth noting for anyone comparing against the #799 numbers: the `test_gfm_bare_urls_emails_and_schemes_cannot_autolink` failure I reported there is **gone** — #802 gated it behind `importorskip(minversion="3.0.0")`, so my ambient markdown-it-py 2.2.0 now skips visibly instead of failing (3 → 4 skipped). This branch runs clean.


Imbad0202 (changes requested):
Thanks for the fast follow-up on this one. The interior-balance insight and the
mutation verification are the right instincts, and the oracle extension is clean.
One finding needs a fix before merge; I have verified every number below locally.

**P1: `’` and `”` in non-wrapper roles defeat the balance scan.**

The interior scan treats every `’` as a closer of `‘`, but `’` is also the
apostrophe in embedded English. Verified counterexample:

`《Alzheimer’s病中ＰｒｏＥＸＣ表达》` vs `Alzheimer’s病中ProEXC表达`

On main both sides normalize identically and match. On this branch the lone `’`
blocks the outer strip: `exact_normalized_title` is False and `_similarity` is
0.6818, below the 0.70 floor. That takes out both resolver paths at once (the
DOI-keyed ratio gate and the title-fallback exact gate), and the CJK client's
exact-or-bust DOI check fails with them, so a correct DOI can be reported as a
mismatch. This is the failure class #798 just repaired, so I can't accept it
under the issue's conservative-direction blessing: that blessing covers titles
whose outer marks are genuinely not one pair, not titles the outer pair does
enclose.

Suggested direction (yours to choose): scope the interior balance scan to the
outer pair's own family. Every case in your behavior table survives
(`《红楼梦》与《金瓶梅》` still trips on the stray `》`, `“研究”与“实践”` on the
stray `”`, nested `「」` still unwraps), while an interior apostrophe no longer
vetoes a genuine `《…》` wrap. Please add wrapped-vs-bare regressions for the
apostrophe shapes (`’s`, `’98`).

Three advisories, cheap to take in the same pass:

- The wrong-family-closer branch (`stack.pop()` mismatch) has no discriminating
  test: a mutant that accepts any closer after any opener passes the current
  suite. Pin it under whatever semantics the P1 fix lands on (note the
  interaction: family-scoped balance changes what `《「研究』》` should do).
- The module docstring still says "Behaviorally equivalent to the implementation
  this was promoted from". That was true until this PR changed the behavior; it
  needs a historical qualifier.
- The CHANGELOG's non-CJK invariance claim is broader than the code:
  `_cn_titles_match` calls `normalize_cn_title` without a `has_cjk` gate, so
  `《O’Connor Study》` vs its bare form flips from match to non-match on the
  client path, and the "`《》` never matches" line only holds on the gated paths
  (`exact_normalized_title("《》", "《》")` is still True). Either gate the client
  call or narrow the claim.

Happy to merge once the P1 lands. The rest of the PR is solid and I'd like it in
the next patch release.

akshath-raj:
Thanks — the P1 is a real regression and you were right to block on it. All four items are in `01fdbd9`.

## P1: interior scan is now scoped to the outer pair's own family

Reproduced your counterexample against `main` before touching anything: `《Alzheimer’s病中ＰｒｏＥＸＣ表达》` vs `Alzheimer’s病中ProEXC表达` went from exact **True / ratio 1.0** on `main` to **False / 0.6818** on my branch. Below the 0.70 floor, so both resolver paths fail together — exactly the #798 failure class, and I agree the conservative-direction blessing doesn't stretch to a title whose outer pair genuinely does enclose.

Took your suggested direction. A `《…》` wrap now tracks only `《`/`》` and is blind to quote marks. The argument that this loses nothing: any mark that can orphan the **outer** pair is by definition of that pair's own family, so the other five families were never buying the check anything — they were only ever able to produce false vetoes like this one. Every row of the behavior table survives, and the two lookup maps the family-blind version needed are now dead and removed.

Regressions added for both apostrophe shapes (`’s` and `’98`), asserting normalized equality, `exact_normalized_title`, and ratio 1.0.

On the interaction you flagged: `《「研究』》` → `「研究』`. The outer `《》` does enclose one unit, so it strips; the mismatched inner quotes survive as content, which reads correct to me since they're exactly what distinguishes it from `《研究》`. Pinned explicitly rather than left implicit.

## Advisory 1: depth-0 stray-closer branch

Pinned — but my first attempt at this test **did not discriminate**, and the reason is worth recording because it would trap the next person.

Clamping (`depth = max(0, depth - 1)`) absorbs one closer, so any title with equal opener/closer counts — every natural case, `《红楼梦》与《金瓶梅》` included — still ends at depth 1 and gets refused by the *trailing* check anyway. Discriminating requires interior closers to outnumber openers by exactly the clamp count, which no natural title shape produces. So that test uses a documented synthetic (`《》研究《实践》》`) asserted against the helper directly, with the arithmetic written into the docstring so it doesn't get "simplified" back into a natural-looking title later.

All three branches are now individually mutation-verified: clamp-instead-of-refuse, drop-the-trailing-check, and family-blind (the P1 regression itself).

## Advisory 2: docstring

Qualified as historical — true of the #798/#799 promotion, false since #800 deliberately changed this behavior.

## Advisory 3: invariance claim narrowed

You're right on both halves. One note: your `《O’Connor Study》` example now **matches again**, because the apostrophe was the cause and the P1 fix resolved it. The underlying hole is real regardless, so I narrowed rather than leaning on that.

The CHANGELOG now says: the claim holds on the two `has_cjk`-gated paths; `_cn_titles_match` is ungated so a mark-carrying Han-free title can change verdict there (`《Hamlet》and《Macbeth》` vs its pre-mangled form — verified to flip); and the empty-wrapper guarantee is a `_cjk_titles_match` property, with `exact_normalized_title("《》", "《》")` still True through the base branch as before.

I chose narrowing over gating the client call: gating touches #799's contract and the path is DOI-keyed and Chinese-corpus-only in practice, so the narrowing is honest and the smaller change. Happy to gate it instead if you'd rather — say the word. Both narrowings are pinned by tests so the prose can't drift from the code.

## Verification

11 new tests. Full suite: **9270 passed, 4 skipped, 0 failed.** PR body updated throughout, including the behavior table and the stale "all six wrapper pairs" description.

One thing I'll own: my mutation testing in the first round checked that the interior scan was load-bearing, but never asked whether it was too *aggressive* — every case I tried had wrapper marks in wrapper roles. The apostrophe is the same codepoint in a different role, which is the case that class of testing structurally misses. Noted for next time.

Imbad0202 (approved):
Approved. Merging once the two remaining CI jobs go green.

On the question you left open: keep the narrowing. Gating the `_cn_titles_match` call would reach into #799's contract for a path that in practice only sees the Chinese corpus, and both narrowings are already pinned by tests, so the prose cannot drift. Not worth the larger change.

I re-verified the branch locally before approving: the apostrophe pair matches again, `《红楼梦》与《金瓶梅》` keeps its marks, and `test_text_similarity.py` passes clean.

The closing note on what first-round mutation testing structurally misses is the kind of record that saves the next person who touches this file. Thank you for a careful round.

PR #816, opened 2026-09-02
feat: Firecrawl research-index client + API protocol reference

rakshith48:
## Summary

Firecrawl's **research index** was not reachable from this repo at all (no `firecrawl` reference anywhere before this branch). This adds it in the shape the repo already uses for bibliographic upstreams: a standalone client + a mirrored protocol reference + a test suite.

**Why it's worth having.** The four gate resolvers (S2 / OpenAlex / Crossref / arXiv) key their *ID* paths on a DOI or an arXiv ID. A conference paper without a registered DOI, a PubMed-only record whose DOI the author omitted, or a preprint cited by title therefore reduces to `unresolvable` — the same state a fabricated reference produces. Firecrawl's research index keys on its own canonical `paperId` plus `arxiv:` / `pmid:` / `pmcid:`, so it can put an ID-keyed answer under some of those citations. It also exposes paper **full-text passages**, which no sibling resolver does.

**New files**
- `scripts/firecrawl_client.py` — `FirecrawlResearchClient` with `paper_id_lookup` (ID-keyed metadata under the siblings' 0.70 title cross-check / ID_MISMATCH pattern), `title_search` (ranked search behind the #431 exact-title-or-bust gate), `paper_passages` (full-text passages, advisory).
- `deep-research/references/firecrawl_api_protocol.md` — verified live examples, accepted id grammar, degradation table, and a "Where this stops" section.
- `scripts/test_firecrawl_client.py` — 54 tests, fully synthetic inline bodies, zero live network.

**Three one-line insertions into existing files** (no refactoring anywhere):
- `docs/DATA_FLOWS.md` — a network-touchpoint row (required by the DF-1 gate in `check_data_flows.py`).
- `scripts/_ci_pytest_manifest.toml` — one `[[pytest]]` entry.
- `CHANGELOG.md` — one `[Unreleased] / Added` entry.

**Transport: REST over stdlib `urllib`, no vendor SDK.** Every sibling client (`arxiv_client.py`, `openalex_client.py`, `crossref_client.py`, `semantic_scholar_client.py`, `chinese_literature_client.py`) is stdlib-only with zero runtime dependencies, so adding the Firecrawl SDK would have been the awkward choice. The v2 endpoint paths were verified against the live API.

**Redirects are refused.** CPython's `HTTPRedirectHandler` copies every header except content-length/content-type onto the redirected request, so a followed cross-origin `Location` would carry the optional `Authorization` header to a third-party host. Requests route through an opener that raises on every redirect hop — the same posture `chinese_literature_client.py` takes toward off-allowlist hops. This is the one deliberate divergence from the sibling clients: they send no `Authorization` header, so a followed redirect cannot leak a credential for them.

### Standalone by design

No `verification_gate`, `citation_verification_summary`, triangulation-matrix, schema, or `degradation_registry.json` changes — the same boundary `chinese_literature_client.py` (#595) holds. This is not deferral. A **ranked semantic** index structurally cannot supply the triangulation matrix's `*_unmatched` boolean: the title path always returns nearest neighbours, so "search returned rows" is never existence evidence. Measured on a deliberately fabricated title, the index returned three real, unrelated papers (0.207 / 0.192 / 0.187, against 0.989 for a genuine exact title). The client therefore exposes **no `*_unmatched` signal**, and `title_search` keeps the exact-title gate for exactly that reason.

Three further upstream properties, each measured and documented rather than papered over:
- **The index is not papers-only.** `web:`-namespaced records share the paper response shape (`query=test` returned a Merriam-Webster dictionary entry). They're dropped before scoring; an id-less record is also dropped, so a future non-paper namespace fails closed.
- **No venue publication year exists** (only index-side `createdDate` / `updateDate`), so the siblings' `+0.05` matching-year tiebreaker is deliberately **not** reimplemented — reading `createdDate` as a publication year would misdate every published-after-preprint work in a corpus.
- **An out-of-grammar id raises rather than missing.** A DOI's `/` leaves the path route and answers 404 HTML; a DOI-keyed caller reading `None` as "the index does not hold this work" is the one confusion that must not be possible where the gate distinguishes `unresolvable` from `false`. The same rule applies to schema drift: a `success: true` body without a well-formed `paper` raises instead of collapsing to `None`.

`POST /v2/search` with `categories: ["research"]` is documented as **deliberately not wired**: it returns web pages, not index records, so nothing on it is ID-keyed, and it falls inside the bounded browser-fallback role the #495 retrieval-order boundary already governs. Wiring a general web-search channel into the literature path is a behavior change that belongs in an issue first.

The protocol doc also records, against this PR's own novelty, that **the PubMed part of the gap has a cheaper alternative**: S2's API accepts `PMID:` / `PMCID:` prefixes on the `/paper/{id}` route the gate already calls, and `semantic_scholar_client.py` simply doesn't use them. Extending that client would be a smaller change and would keep the answer inside a resolver the gate already trusts. This index's distinct contributions are the unified arXiv-plus-biomedical key space and the passages endpoint.

## Eval impact

`scripts/firecrawl_client.py` is not in the `eval-harness.yml` Delta 3 path filter (which names the gate-wired resolver clients specifically), and nothing in this branch touches scoring, generation, or gold sets.

> No eval impact.

## Platform port?

> Not a platform port.

## Verification

**Live run against the real API**, confirmed by server-generated response headers (`x-request-id`, `x-response-time`) captured around the client's own transport:
- `paper_id_lookup("arxiv:1706.03762", "Attention Is All You Need")` → matched record
- ID_MISMATCH path (unrelated expected title) → `None`; 404 on a bogus id → `None`
- `title_search("Attention is all you need")` → exact hit; fabricated title → `None`
- `paper_passages(...)` → real passage text with retrieval scores
- **Keyless** path verified live (no `Authorization` header sent, real response returned), which is what lets this client keep the gate's key-free footing
- Redirect refusal verified end-to-end against a local redirecting server: with the fix, the `Bearer` key never reaches the redirect target

**Repo gates — all passing locally** (with `requirements-dev.txt` installed, `PYTHONPATH=.`, `GITHUB_DEFAULT_BRANCH=main`, as `pytest.yml` runs them):
- `pytest scripts/ tests/` (full suite) → passed
- `pytest scripts/test_firecrawl_client.py` → **54 passed**
- `check_data_flows.py` → OK (DF-1..DF-3)
- `check_ci_pytest_manifest.py` → OK, and `run_ci_pytest_manifest.py --id firecrawl-research-index-client` → passed
- `check_spec_consistency.py`, `check_version_consistency.py` → OK
- `py_compile` clean on both new modules (no Python linter config exists in the repo, so the lint scripts above are the lint surface)

`test-count-monotonic` should see **+54**, never a reduction.

## Notes for review

- **The standalone boundary.** The judgement that this must not feed the triangulation gate is argued in "Where this stops". I believe it's the correct call, but it is a maintainer's call.
- **The chosen surface.** `paper_passages` is the most opinionated piece: it returns evidence for a human to read and never a verdict, deliberately. If you'd rather it not ship until there's a consuming agent, it can be dropped without touching the other two methods.
- **`FIRECRAWL_API_KEY` as a CI secret is NOT required**, and I'd suggest not adding it: the tests use synthetic bodies and make zero network calls, and the client works keyless. A key only raises the rate limit.
- **Keyless behaviour from GitHub Actions runners is unverified.** Firecrawl's keyless tier can be IP-reputation gated, so a runner could behave differently. Nothing in CI depends on it.
- **Retry set deliberately unchanged.** Firecrawl documents 408/5xx as retryable, but all five sibling resolvers raise on 5xx with no retry as a load-bearing degradation contract. Diverging here should be a cross-client decision, not this client quietly differing.
- **No README changes**, deliberately: the comparable standalone client (#595) isn't in the READMEs either, and the user-facing surface for network touchpoints is `docs/DATA_FLOWS.md` (already linked from the README). Happy to add a README section if wanted — that would mean the five translated READMEs moving together per CONTRIBUTING.
- **`/papers/{id}/similar`** is documented but not implemented; the `k` param form used elsewhere returned HTTP 400, so a future caller must re-derive its query grammar.
- **No issue number.** CONTRIBUTING puts "new reference files" under *requires maintainer review* rather than *open an issue first*, so this comes straight as a PR. The CHANGELOG entry carries no `#N` yet; happy to add this PR's number (or an issue you'd prefer to track it under) on request.

[skip-closes-check] — additive standalone client with no pre-existing issue; CONTRIBUTING classifies new reference files as maintainer-review, not issue-first.

## Checklist


Imbad0202:
Thanks for the careful work here, and for measuring upstream behavior rather than assuming it. The client is well built and the tests are thorough.

We are going to decline this for the core tree, and the reason is scope rather than quality. ARS keeps a small set of resolvers that are public registries or public indexes (Crossref, OpenAlex, Semantic Scholar, arXiv). Firecrawl's research index is a closed cloud service from a single vendor, and its coverage (arXiv plus PubMed, bioRxiv, medRxiv) is already reachable through those four. That leaves the passages endpoint as the one distinct capability, and we do not have a consuming agent for it. Maintaining a vendor client without a consumer is a cost we would rather not take on in core.

If you publish this as a standalone package, we would be glad to list it under THIRD_PARTY.md. Two things you may want to fix before that, both verified live today: the index does accept `doi:` ids when the DOI is percent-encoded (`quote(id, safe='')`), so the "DOIs are not a key" statement in the protocol doc is an encoding artifact rather than an upstream property; and a route-level 404 (the open-source proxy returns an empty 404 when `RESEARCH_PROXY_URL` is unset) currently reads as a record miss, while a real miss carries a JSON body with `code: NOT_FOUND`. A third, smaller one: an ID-path record with a missing or non-string `title` currently becomes an ID_MISMATCH instead of a degradation.

Thanks again for bringing this to us.

Issue #834, opened 2026-09-07
Socratic non-convergence path F6 still preselects "the most promising direction" and auto-ends at round 15, contradicting the #735 non-ranking boundary and the #490 turn-cap authority


Imbad0202:
## Problem

#735 (closed 2026-08-15) aligned the Socratic non-convergence boundary across surfaces: the default fallback "summarizes only directions the scholar has already expressed", and "no fallback presents a ranked or preselected option". At `v3.21.2` one failure-path surface still does both, and a second surface carries a turn cap that the agent file superseded.

### 1. F6 still preselects a direction and narrows scope

`deep-research/references/failure_paths.md` § F6 (Socratic Dialogue Does Not Converge), at `v3.21.2`:

- User notification option (A): "Continue the Socratic dialogue, but focus on **[the most promising direction]** you just mentioned?"
- Handling step 2: "**Identify the 1-2 directions with the most convergence potential**"
- Recovery path: "Continue with focus → **restrict discussion scope**, converge within 5 rounds"

The directions are the researcher's own, so this is not candidate *generation* (the #735 fixture that proves unsuccessful rounds do not authorize generation passes). It is the AI ranking the researcher's directions by "convergence potential" and pre-filling one as "most promising", which is exactly the preselected option #735's fourth acceptance item forbids and the POSITIONING "Rejected mechanisms" list rules out ("propose, substitute, rank, expand, or select"). The narrowing prescription ("restrict discussion scope") also runs against the sequential-narrowing concern that the ideation-diversity design (#659) exists to measure. `deep-research/references/socratic_mode_protocol.md` § Dialogue Management Rules points at F6 as the reference for the non-convergence path, so the contradiction is live, not orphaned.

### 2. Two surfaces still say "round 15 → end"

- `failure_paths.md` § F6 step 4: "If user chooses to continue but still hasn't converged by round 15 → auto-compile + end"
- `socratic_mode_protocol.md` § Dialogue Management Rules: "If dialogue exceeds 15 rounds -> automatically compile INSIGHTs and end"

`deep-research/agents/socratic_mentor_agent.md` § Auto-End Conditions (Precise) is the single authority since #490 and says 40 rounds (goal-oriented) / 60 (exploratory), with stagnation detection at 10 / 15. The two reference files were not updated when the authority moved.

## Proposed correction

[...]

Issue #735, opened 2026-08-15
Align the Socratic research-question generation boundary across all surfaces


Imbad0202:
## Problem

The declared Socratic boundary says ARS does not originate, rank, or choose the scholar's research question, but the non-convergence fallback can generate candidate questions and ask the user to select one. That is an invisible mode change.

Reproducible surfaces include `POSITIONING.md`, `deep-research/SKILL.md`, `deep-research/agents/research_question_agent.md`, and `deep-research/references/failure_paths.md`.

## Required correction

- The default non-convergence fallback summarizes only directions the scholar has already expressed and offers continued dialogue or literature exploration.
- It must not generate, rank, recommend, or select candidate research questions.
- If a user explicitly asks ARS to propose candidates, the workflow must obtain that instruction and render a visible receipt that the session has left the non-generative Socratic contract. Generated candidates remain proposals for author adjudication, never an inferred author choice.
- All maintained command, skill, agent, and failure-path surfaces use the same rule.

## Acceptance

- [ ] A static consistency test pins the default no-generation fallback across all named surfaces.
- [ ] A fixture proves five or more unsuccessful rounds do not authorize candidate generation.
- [ ] A fixture proves explicit user-requested generation produces a visible mode-transition receipt.
- [ ] No fallback presents a ranked or preselected option.
- [ ] Documentation states that this correction removes a contract contradiction; it does not establish improved ideational breadth.

## Non-goals

No branch ledger, creativity score, or effectiveness claim. Breadth measurement remains in #659.

Parent: #734.

PR #836, opened 2026-09-08
Fix: Socratic non-convergence path F6 still preselects "the most promising 

goransh-walia:
This PR addresses #834.

Socratic non-convergence path F6 still preselects "the most promising direction" and auto-ends at round 15, contradicting the #735 non-ranking boundary and the #490 turn-cap authority

### DIFF
--- a/deep-research/references/failure_paths.md
+++ b/deep-research/references/failure_paths.md
@@ -1,3 +1,4 @@
+---
 # Failure Paths — Research Pipeline Failure Path Map
 
 ## Overview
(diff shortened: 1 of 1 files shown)


Imbad0202:
Thanks for picking up #834. I'm closing this in favor of #840, which carries the whole scope of the issue: the F6 rewrite, the matching line in `socratic_mode_protocol.md`, a static test that fails on the pre-fix bytes, and the CHANGELOG entry.

Two things in this diff would also have blocked a merge: the `---` added at line 1 opens a YAML frontmatter block that swallows the title, and the file's final newline was dropped.

If you'd like to contribute again, a short note on the issue before coding, saying what you plan to change, lets me flag things like byte pins and sibling files early. Happy to answer questions there.

PR #838, opened 2026-09-08
Fix: Socratic non-convergence path F6 still preselects "the most promising 

goransh-walia:
This PR addresses #834.

Socratic non-convergence path F6 still preselects "the most promising direction" and auto-ends at round 15, contradicting the #735 non-ranking boundary and the #490 turn-cap authority


Imbad0202:
Same as #836, closing in favor of #840.

Issue #833, opened 2026-09-07
[Gartenberg sub D] Docs: fourth human-in-the-loop anchor, volume non-goal in POSITIONING, cognitive-surrender and hedging notes


Imbad0202:
> Sub-issue of [META] Gartenberg et al. 2026 (#829)

#### Anchor

- Editorial abstract and §5: "the current state of AI tools, amplified by existing publish-or-perish incentives, appears to be pushing the system toward an equilibrium of more rather than better research"; §5.1 on "cognitive surrender" (Shaw & Nave 2026) vs. human-first use; §5.2 on craft erosion (Bechky & Davis 2025) and on measuring progress across model generations rather than assuming it.
- ARS files: `README.md` (motivation section with three anchors), `POSITIONING.md` ("Rejected mechanisms"), `shared/collaboration_depth_rubric.md` (Wang & Zhang 2026), `shared/references/claim_strength_ladder.md`, `docs/PERFORMANCE.md`.

#### Proposed change (docs only)

1. README motivation: add the editorial as the fourth anchor, cited as design rationale, with its scope stated (one journal, observational, aggregate). One paragraph, same register as the Zhao and Ren paragraphs.
2. POSITIONING "Rejected mechanisms": record the volume non-goal explicitly: ARS never batch-generates manuscripts, never drives multiple submissions from one run, and treats time-to-draft as a resource cost, not an outcome. The editorial is the external reason to say this out loud.
3. `collaboration_depth_rubric.md`: a related-construct note linking Cognitive Vigilance to "cognitive surrender" (Shaw & Nave 2026, preprint) as the failure the vigilance dimension is meant to make visible; no rubric change.
4. `claim_strength_ladder.md`: one line of external evidence that AI-written text hedges less (editorial Fig. 8), i.e. the ladder's guarded direction is the empirically observed drift direction.

#### Acceptance

- [ ] Four docs touched, each with the citation and a one-sentence scope caveat
- [ ] No new effectiveness claim anywhere; the release-notes lint and the anchored claim sentences stay green

#### Non-goals

- No numbers from the editorial presented as if they were about ARS.

Imbad0202:
Item 4 (the `claim_strength_ladder.md` line) is deferred to #830 by maintainer decision.

[...]

PR #837, opened 2026-09-08
Fix: [Gartenberg sub D] Docs: fourth human-in-the-loop anchor, volume non-g

goransh-walia:
This PR addresses #833.

[Gartenberg sub D] Docs: fourth human-in-the-loop anchor, volume non-goal in POSITIONING, cognitive-surrender and hedging notes

Generated with AI assistance and validated against the original
file before submission (syntax check + change-scope check).
Please review carefully — happy to adjust based on feedback.

PR #839, opened 2026-09-08
Fix: [Gartenberg sub D] Docs: fourth human-in-the-loop anchor, volume non-g

goransh-walia:
This PR addresses #833.

[Gartenberg sub D] Docs: fourth human-in-the-loop anchor, volume non-goal in POSITIONING, cognitive-surrender and hedging notes


Imbad0202:
Same as #837, closing in favor of #841. Note that `claim_strength_ladder.md` is byte-pinned by the revision-claim-drift suite, so an edit there has to re-pin the suite in the same commit.

PR #844, opened 2026-09-08
fix: make /ars-mark-read work on Windows (#843)

### LINKED ISSUE #843: fix: /ars-mark-read fails on Windows because of POSIX-only fcntl import

dajiaohuang:
## Summary

Fixes #843. scripts/ars_mark_read.py imported POSIX-only fcntl at module load, so the documented /ars-mark-read CLI failed on Windows before parsing arguments.

The patch keeps fcntl.flock on POSIX and adds a stdlib msvcrt.locking fallback on Windows. It preserves the existing bounded retry loop, exclusive lock, persistent lock file, and clean release behavior.

## Verification

- pytest -q tests/test_mark_read_args.py — 4 passed on Windows/Python 3.13.
- Before the patch, the same focused suite had 4 collection/runtime failures with ModuleNotFoundError: No module named 'fcntl'.
- Full pytest -q remains blocked by pre-existing Windows-only fcntl imports in other scripts and missing ruamel.yaml; those are recorded in the campaign audit.

This is a narrow external-contributor PR; no maintainer approval or merge is implied.

### DIFF
--- a/scripts/ars_mark_read.py
+++ b/scripts/ars_mark_read.py
@@ -157,8 +157,6 @@ def _ledger_lock(
 
     acquired = False
     try:
-        if fcntl is None and os.fstat(fd).st_size == 0:
-            os.write(fd, b"\0")
         os.lseek(fd, 0, os.SEEK_SET)
         deadline = time.monotonic() + timeout
         while True:
--- a/scripts/ars_mark_read.py
+++ b/scripts/ars_mark_read.py
@@ -37,7 +37,6 @@ from __future__ import annotations
 
 import argparse
 import errno
-import fcntl
 import os
 import sys
 import tempfile
@@ -49,6 +48,12 @@ from pathlib import Path
 from typing import Any, Iterator
 
 import yaml
+
+try:
+    import fcntl
+except ModuleNotFoundError:  # pragma: no cover - exercised on Windows
+    fcntl = None  # type: ignore[assignment]
+    import msvcrt
 
 try:
     from scripts.human_read_attestation_resolver import (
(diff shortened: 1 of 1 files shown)


Imbad0202 (changes requested):
Thanks for the reproduction and the focused patch. The fcntl/msvcrt split and the two lock helpers look right, and the POSIX path is unchanged.

One change before merge, at `scripts/ars_mark_read.py:160-161`: please drop the empty-file check and the NUL write.

- `msvcrt.locking` can lock a region beyond EOF (Python docs: "may continue beyond the end of the file"), so the lock file does not need a first byte.
- The write sits outside the retry loop. If two first-time writers both see size 0, the second `os.write` fails with a raw OSError instead of waiting on the lock. Narrow window, but the fix is a two-line deletion.

The `os.lseek(fd, 0, os.SEEK_SET)` line can stay.

Everything else checks out: `LK_NBLCK` contention raises `EACCES`, which the existing loop already retries. Reviewed with an independent cross-model pass (GPT-6 Astra) plus a manual read; both landed on the same single finding.

Once that is in, I will approve the CI run and merge. The five remaining `import fcntl` sites (adjudication_activity, inquiry_branch_ledger, review_criteria_binding and their tests) are out of scope here; I will open a follow-up issue for them.

dajiaohuang:
Addressed the requested Windows lock change in 6f1e4b0: removed the empty-file check and the pre-write of a NUL byte. The lock path now lets msvcrt.locking acquire the byte at EOF as requested. Focused validation: python -m pytest -q tests/test_mark_read_args.py (4 passed).

Imbad0202 (approved):
6f1e4b0 is exactly the two-line deletion requested. CI green after approval (pytest, Spec Consistency, hygiene, command invariants, test-count monotonic). I also repaired the PR body rendering (literal \n and \f escapes had turned fcntl into a form-feed) so the closes-issue check could find "Fixes #843"; wording unchanged. Merging. The remaining five fcntl import sites are tracked in #845.

Issue #845, opened 2026-09-09
Windows portability: remaining `fcntl` imports after #844


Imbad0202:
#844 fixes the documented `/ars-mark-read` CLI. Five more files still import the POSIX-only `fcntl` module at load time and fail on Windows with `ModuleNotFoundError`: `scripts/adjudication_activity.py`, `scripts/inquiry_branch_ledger.py`, `scripts/review_criteria_binding.py`, and their two test files.

Rather than repeating the #844 try/except pattern in each file, extract one shared lock helper (e.g. `scripts/file_lock.py`) and route all six sites through it, `ars_mark_read.py` included. Two semantic gaps need a decision, not a mechanical port:

- `adjudication_activity.py:366` takes a shared lock (`LOCK_SH`). `msvcrt.locking` has no shared mode; the Windows backend would need to either downgrade to exclusive or document the difference.
- `review_criteria_binding.py:505` takes a blocking exclusive lock. `msvcrt.LK_LOCK` retries 10 times then raises `EDEADLOCK`, so the helper should implement its own bounded wait on Windows.

Acceptance: full `pytest -q` passes on Windows/Python 3.13 (the #843 reporter's environment), verified manually. Windows stays best-effort: no `windows-latest` CI job is added, and the SETUP docs should say so where they describe platform support.

Refs #843, #844.

Imbad0202:
@dajiaohuang Since you have the Windows/Python 3.13 environment that surfaced #843, would you be interested in taking this one after #844 lands? The shape I have in mind is a single shared lock helper that all six sites route through, rather than repeating the try/except per file. The two semantic gaps in the issue body (shared lock, blocking lock) are the parts that need a design call; happy to discuss them here before any code.

No pressure either way. If you would rather not, I will pick it up after #844 merges and ask you to verify the Windows run.

Imbad0202:
PR #847 implements this: one shared helper (`scripts/file_lock.py`, `fcntl` on POSIX / `msvcrt` byte-0 lock on Windows) routes all four scripts, with the two semantic gaps decided per site rather than hidden:

- adjudication store: shared reads degrade to an exclusive lock with a 5 s bounded wait on Windows; writers stay non-waiting.
- review-criteria manifest: still blocks indefinitely on POSIX; capped at 30 s on Windows, reported as `BindingError`.
- inquiry branch ledger: keeps refusing non-POSIX hosts.

[...]

PR #847, opened 2026-09-09
fix: shared file-lock helper with msvcrt backend for the remaining fcntl sites (#845)

Imbad0202:
## Summary

Refs #845 (follow-up to #843 / #844). One shared advisory file-lock helper, `scripts/file_lock.py`, replaces the per-file `fcntl` usage in `adjudication_activity.py`, `inquiry_branch_ledger.py`, `review_criteria_binding.py`, and `ars_mark_read.py`, with an `msvcrt` backend for Windows. The POSIX lock sequences are unchanged.

Per-site Windows decisions (maintainer decisions, not mechanical ports):

- **adjudication store**: shared read locks degrade to exclusive with a 5 s bounded wait; writers keep the non-waiting exclusive lock.
- **review-criteria manifest**: blocks indefinitely on POSIX as before; capped at 30 s on Windows (`WINDOWS_BLOCKING_WAIT_SECONDS`), surfacing as `BindingError`.
- **inquiry branch ledger**: keeps refusing non-POSIX hosts (checks `file_lock.BACKEND`).

Also fixed on the way: two `finally` blocks released a lock they had not acquired (no-op under `flock`, `EACCES` under `msvcrt` that would have masked the real error); both now use `file_lock.held()`. One POSIX-visible textual change: the adjudication contention message names `LockTimeout` instead of `BlockingIOError` (`ERROR:LOCK` and exit status unchanged).

## Verification

- `scripts/test_file_lock.py` (CI manifest id `845-shared-file-lock`): real `fcntl` backend plus `tests/fake_msvcrt.py` for the Windows branch, plus a subprocess test that blocks `fcntl`, imports all four consumers, and exercises each site's Windows decision.
- Full `pytest -q` on macOS: 9511 passed / 3 skipped (before the /simplify refactor); the five affected test files re-run green after it (244 passed). CI runs the full suite.
- Cross-model review (GPT-6 Astra, xhigh, read-only): round 1 found 1 P1 (interrupted retries could pass the deadline) + 2 P2, all closed; round 2 verified the closures with no new findings. Security review: no findings.
- **Not verified on real Windows.** No Windows CI job is added (issue decision). @dajiaohuang, if you can run `python -m pytest -q` on your Windows/Python 3.13 environment against this branch, that is the acceptance step the issue asks for; I will keep #845 open until then.

## Docs

`docs/SETUP.md` / `SETUP.zh-TW.md` gain a platform-support paragraph; CHANGELOG `[Unreleased] › Fixed` entry added.

[skip-closes-check] Refs #845 only: the issue's acceptance criterion (full pytest on Windows) needs a manual run I cannot perform.

https://claude.ai/code/session_0131cZMWBPPeEFiqgEPFZ3X2


Imbad0202:
@dajiaohuang this is the follow-up to your #844. I am holding it open for your review rather than merging on my own timeline.

Two things would help most, in this order:

1. **A read of the two Windows decisions in the description.** The adjudication reader wait (5 s, exclusive fallback) and the 30 s cap on the review-criteria manifest lock are judgement calls, not derived from the msvcrt contract. If either looks wrong from the Windows side, say so here and we can change it before merge.
2. **The Windows run.** `git fetch origin pull/847/head:pr-847 && git checkout pr-847`, then `python -m pytest -q`. The tail of the output is enough. `tests/test_file_lock.py` exercises the msvcrt branch through a fake on other platforms, so your run is the only real-platform evidence this change will have.

No deadline on either. If you would rather not review, a one-line reply is fine and I will merge with the SETUP note that Windows is unverified.

dajiaohuang:
I ran the Windows validation on head 5a8505029 using Python 3.13. The targeted lock-path suite passes: pytest -q scripts/test_file_lock.py → 14 passed, 1 skipped. A full pytest -q collection is blocked by the checkout environment missing 
uamel.yaml in three unrelated corpus-schema tests; no source changes were made. The 5 s adjudication-reader wait and 30 s manifest-lock cap look reasonable as bounded operational defaults, with the tests covering the Windows fallback behavior.

